PR #15, opened 2025-12-10
RUI-167: Publish using trusted publishing

baseilos:

Migrates publishing to https://docs.npmjs.com/trusted-publishers#step-2-configure-your-cicd-workflow

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,6 +5,10 @@ on:
     branches:
       - main
 
+permissions:
+  id-token: write
+  contents: write
+
 jobs:
   publish:
     name: Publish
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -5,6 +5,10 @@ on:
     branches:
       - main
 
+permissions:
+  id-token: write
+  contents: write
```
mad-raccoon, line 10: I guess there is still one line missing here

mad-raccoon: id-token: write
  contents: write
  pull-requests: write

baseilos: In don't think `pull-requests` write are needed actually as it is used to manipulate with pull request itself which we don't need. Let's try it without it.

mad-raccoon (approved):

PR #66, opened 2026-02-02
EM-2621: Fix critical vulnerabilities

ako-hk:

This is to update the sdk to fix the critical vulnerabilities that we have. 

# Main changes

Updated `SDK-core` and `SDK-react` to their latest versions  
# Test evidence
From: 6 vulnerabilities (1 moderate, 2 high, 3 critical)
To: 2 vulnerabilities (1 moderate, 1 high)
Unfortunately for the `high` that we is left, we should replace `xlsx` npm package which will go out of scope of this ticket.

## Summary by CodeRabbit

* **Chores**
  * Updated SDK package dependencies to the latest versions for improved stability and performance.

### DIFF
--- a/.changeset/major-poets-argue.md
+++ b/.changeset/major-poets-argue.md
@@ -1,0 +1,5 @@
+---
+'@embeddable.com/remarkable-pro': minor
+---
+
+Update to latest SDK to fix critical vulnerabilities
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@embeddable.com/remarkable-pro",
-  "version": "0.0.22",
+  "version": "0.0.25",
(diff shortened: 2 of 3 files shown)


mad-raccoon (approved):

.changeset/major-poets-argue.md
mad-raccoon, line 1: i think this didnt need to go as a minor - since there were no breaking changes correct ?

mad-raccoon: (no need to rollback - just trying to understand if was really needed a minor, or if a patch would be enough for future reference)

ako-hk: That's right.
It still could be case by case for this one of course it could be patch

PR #78, opened 2026-02-09
[EXAMPLE ONLY DO NOT MERGE] Example of user-defined custom hook in theme

cwbuecheler:
This shows one way of implementing a custom hook that the user could define on the theme which would allow them access to the props for any given chart, and they could then manipulate those props as they like and return them.

Apologies, the typing for TS compatibility is a hack job, but speed's of the essence in this case! :grin:

This is related to an issue NOQ are having where they need to hit an external API, get some stuff from it, and then use that stuff to manipulate `props.results.data` returned by `loadData` before doing anything in the chart. They have to do this for virtually every chart, with the results from the external API being different for each individual user and dataset, meaning potentially tens of thousands of different results.

## Release Notes

* **New Features**
  * Added theme-level customization capability to enable dynamic prop transformation for charts.
  * Implemented data augmentation support for chart visualization.

* **Refactor**
  * Restructured chart prop flow to utilize theme-based customization system.

### DIFF
--- a/embeddable.theme.ts
+++ b/embeddable.theme.ts
@@ -1,8 +1,35 @@
-import { Theme } from './src/theme/theme.types';
+import { BaseChartTypes, Theme } from './src/theme/theme.types';
 import { remarkableTheme } from './src/theme/theme.constants';
+import { defineTheme } from '@embeddable.com/core';
 
 const themeProvider = (): Theme => {
-  return remarkableTheme;
+  const theme = defineTheme<Theme>(remarkableTheme, {
+    useCustomProps: async <T extends BaseChartTypes>(props: T, chartName: string): Promise<T> => {
+      // Example of a custom props hook that manipulates data from the results prop before it is used in the chart.
+      if (chartName === 'BarChartDefaultPro' && typeof props === 'object' && props !== null) {
+        const data = props.results?.data || [];
+        // Simulate an API request or some data manipulation here. For example, we could filter out certain data points, or add additional data points based on the existing data.
+        const manipulatedData = data.map((row) => {
+          // Example manipulation: add a new field to each row based on existing fields
+          return {
+            ...row,
+            manipulatedValue: row[props.dimension.name] * 2, // Just an example manipulation
+          };
+        });
+        // Return the new props with the manipulated data. The chart will then use this manipulated data instead of the original data.
+        const newProps = {
+          ...props,
+          results: {
+            ...props.results,
+            data: manipulatedData,
+          },
+        };
+        return newProps as T;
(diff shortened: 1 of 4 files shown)


cwbuecheler:
I'm not going to address all the CodeRabbit stuff here because this code is not intended to be used. This whole branch can be ditched once it's determined if this is a project we want to undertake. If so, Rui or whomever will be the one to implement it, not me (most likely). It's just a quick example to help explain what we discussed w/ NOQ as a possbility.

embeddable.theme.ts
```diff
@@ -1,8 +1,35 @@
...
+import { BaseChartTypes, Theme } from './src/theme/theme.types';
 import { remarkableTheme } from './src/theme/theme.constants';
+import { defineTheme } from '@embeddable.com/core';
 
 const themeProvider = (): Theme => {
-  return remarkableTheme;
+  const theme = defineTheme<Theme>(remarkableTheme, {
+    useCustomProps: async <T extends BaseChartTypes>(props: T, chartName: string): Promise<T> => {
+      // Example of a custom props hook that manipulates data from the results prop before it is used in the chart.
+      if (chartName === 'BarChartDefaultPro' && typeof props === 'object' && props !== null) {
```
mad-raccoon, line 9: with this approach all the charts BarChartDefaultPro will go throught the same process even if they dont need the extra data.

in the ideal scenario:

1 - bar chart that get the id-name from the api and them maps (does the api call)
2 - bar chart that gets the state-name from the api and them maps (does the api call)
3 - bar chart normal (doesnt do the api call because is not needed)

cwbuecheler: @mad-raccoon - This example puts the `useCustomProps` function in the R.Pro repo just to show it off in one place, but it wouldn't actually live there. It'd just be in the customers repo. So only customers that wanted to use the hook would define it. Sorry, that wasn't clear.

mad-raccoon: i got it, but this would mean that, if defined, it would apply to all the bar charts, correct? 

my concern here is to avoid this for the barcharts that dont need it

cwbuecheler: Ah, right - yeah, we may need to add some kind of additional way they can say that a specific chart should use the custom props hook. Because yes if they had five bar charts and only needed one of them to run this functionality, that'd be an issue. It'd run for all five.

mad-raccoon: i have 1 idea that might do the trick
let me take some time to test it

cwbuecheler:
Closing this out as we've addressed this user's needs

PR #117, opened 2026-03-10
[RUI-222]: Updates to selects

alicjakujawa:

We need to have better split in selects categories. We also need to have control over apply button in measure/dimension multi selects

# Main changes

* New categories for selects
* Updates to MeasureMultiSelect and DimensionMeasureMultiSelect. 

# Test evidence

https://github.com/user-attachments/assets/09a5ebde-0e37-420c-9b10-7d241710e682

## Release Notes

* **New Features**
  * Dimension and measure multi-select fields now automatically select the first available option on initial load when no options are currently selected.

* **Chores**
  * Reorganized component categories for improved navigation and discoverability.
  * Updated UI library dependency.

### DIFF
--- a/src/components/editors/DimensionOrMeasureMultiSelectFieldPro/index.tsx
+++ b/src/components/editors/DimensionOrMeasureMultiSelectFieldPro/index.tsx
@@ -29,6 +29,7 @@ const DimensionMeasureMultiSelectFieldPro = (props: DimensionMeasureMultiSelectF
   const { title, description, tooltip, placeholder } = resolveI18nProps(props);
 
   const [searchValue, setSearchValue] = useState('');
+  const [pendingValues, setPendingValues] = useState<string[]>([]);
   const hasInitializedSelection = useRef(false);
 
   // Pre-select first option on first load when not clearable (require at least one selection)
@@ -62,12 +63,14 @@ const DimensionMeasureMultiSelectFieldPro = (props: DimensionMeasureMultiSelectF
     <EditorCard title={title} description={description} tooltip={tooltip}>
       <MultiSelectField
         isClearable={clearable}
+        disableApplyButton={!clearable && pendingValues.length === 0}
         isSearchable
         values={currentDimensionAndMeasureNames}
         options={options}
         placeholder={placeholder}
         noOptionsMessage={i18n.t('common.noOptionsFound')}
         onChange={handleChange}
+        onPendingChange={setPendingValues}
         onSearch={setSearchValue}
         avoidCollisions={false}
       />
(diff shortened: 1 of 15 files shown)


src/components/editors/MeasureMultiSelectFieldPro/index.tsx
```diff
@@ -24,6 +24,18 @@ const MeasureMultiSelectFieldPro = (props: MeasureMultiSelectFieldProProps) => {
...
+
+  // Pre-select first option on first load when not clearable (require at least one selection)
+  useEffect(() => {
+    if (clearable) return;
+    const firstOption = measureOptions[0];
+    if (!hasInitializedSelection.current && selectedMeasures.length === 0 && firstOption) {
+      hasInitializedSelection.current = true;
+      onChange([firstOption]);
+    }
+  }, [clearable, measureOptions, selectedMeasures.length, onChange]);
```
mad-raccoon: if i am not mistaken, this can be simplified like this (also this measureOptions[0] could break if the array in empty):

```
useEffect(() => {
  if (clearable) return;
  if (selectedMeasures.length > 0) return;
  if (measureOptions.length === 0) return;

  onChange([measureOptions[0]]);
}, [clearable, selectedMeasures.length, measureOptions, onChange]);
```

mad-raccoon: could you give it a spin and see if it works ?

mad-raccoon (changes requested):

src/components/editors/MeasureMultiSelectFieldPro/index.tsx
alicjakujawa: yes, that works, updated

src/components/editors/DimensionOrMeasureMultiSelectFieldPro/index.tsx
```diff
@@ -29,6 +29,16 @@ const DimensionMeasureMultiSelectFieldPro = (props: DimensionMeasureMultiSelectF
...
+  const [pendingValues, setPendingValues] = useState<string[]>([]);
+
+  useEffect(() => {
+    if (clearable) return;
+    if (selectedDimensionsAndMeasures.length > 0) return;
+    const first = dimensionAndMeasureOptions[0];
+    if (!first) return;
+
+    onChange([first]);
+  }, [clearable, selectedDimensionsAndMeasures.length, dimensionAndMeasureOptions, onChange]);
```
mad-raccoon, line 41: Since we only care about the first option, depending on the entire measureOptions array may cause unnecessary effect runs if the array identity changes. Does it make sense ? 


```suggestion
const firstOption = measureOptions[0];

useEffect(() => {
  if (clearable) return;
  if (selectedMeasures.length > 0) return;
  if (!firstOption) return;

  onChange([firstOption]);
}, [clearable, selectedMeasures.length, firstOption, onChange]);
```

mad-raccoon: this is a nit pick - fell free to ignore

mad-raccoon (approved):

PR #122, opened 2026-03-11
[RUI-246]: Expose fromDate and toDate props

alicjakujawa:

We want to be able to add filters like “up to and including [the to date]”.

# Main changes

Added fromDate, toDate to date pickers.

# Test evidence

https://github.com/user-attachments/assets/59e2232c-c64b-4374-9d5b-c89e3c48a59f

## Summary by CodeRabbit

* **New Features**
  * Date range picker components now expose individual start and end dates separately in change events, providing direct access to selected date boundaries for more flexible date range handling.


src/components/editors/dates/DateRangePickerCustomPro/definition.ts
```diff
@@ -64,9 +74,14 @@ const props = (inputs: Inputs<typeof meta>) => {
...
-    value: range ?? Value.noFilter(),
-  }),
+  onChange: (range: unknown) => {
+    const timeRange = range as { from?: Date; to?: Date } | undefined;
+    return {
+      value: range ?? Value.noFilter(),
+      fromDate: timeRange?.from,
+      toDate: timeRange?.to,
+    };
+  },
```
mad-raccoon:
```suggestion
  onChange: (range: TimeRange) => {
    return {
      value: range ?? Value.noFilter(),
      fromDate: range?.from ?? Value.noFilter(),
      toDate: range?.to ?? Value.noFilter(),
    };
  },
```

we have a type for the range - it might have gotten lost in some refactoring

mad-raccoon: could you test this ? i think you will always need the ?? Value.noFilter() - when the user clears the date range

how you can test:

create a bar chart component with a filter pointing to your fromDate:

Image

after setting a date range, it should show data starting on the fromDate. After you clear the date range, the filter should be reset, and all the data should be returned.

let me know if it is clear.

mad-raccoon (changes requested):

mad-raccoon (approved):

mad-raccoon (commented):
we also need to release a new version

PR #123, opened 2026-03-11
RUI-000: rename bar chart stacked horizontal

mad-raccoon:

Rename BarChartStackedHorizontal and update exports

## Summary by CodeRabbit

* **Chores**
  * Renamed stacked horizontal bar chart component for improved naming consistency. This is a breaking change for API users importing this component.

### DIFF
--- a/.changeset/wide-singers-strive.md
+++ b/.changeset/wide-singers-strive.md
@@ -1,0 +1,5 @@
+---
+'@embeddable.com/remarkable-pro': patch
+---
+
+Rename BarChartStackedHorizontalPro and type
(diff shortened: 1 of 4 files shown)


alicjakujawa (approved):

mikhail-terekhov:
@mad-raccoon isnt this a breaking change? Braking changes are usually released as major, not patch

Non braking change would be exporting old name with the old alias for backward compatibility

mikhail-terekhov:
I see that code rabbit left a similar comment and it was marked as resolved

mad-raccoon:
I marked as solved because currently there is only 1 user extending the components. And he is aware of this change. 

For future breaking changes i Will make sure to use the right versioning

PR #126, opened 2026-03-12
Markdown update

alicjakujawa:

Provided default value to Markdown component

# Main changes

Provided default value to Markdown component

## Summary by CodeRabbit

* **Bug Fixes**
  * The markdown component now properly handles default values when no explicit value is provided, ensuring smooth initialization.

* **Tests**
  * Added test coverage to verify the markdown component functions correctly when initialized without an explicit value.

### DIFF
--- a/src/editors/MarkdownEditor/index.tsx
+++ b/src/editors/MarkdownEditor/index.tsx
@@ -2,12 +2,12 @@ import { FC } from 'react';
 import { MarkdownEditor } from '@embeddable.com/remarkable-ui';
 
 type MarkdownInputProps = {
-  value: string;
+  value?: string;
   onChange: (newValue: string) => void;
 };
 
 const MarkdownInput: FC<MarkdownInputProps> = ({ value, onChange }) => {
-  return <MarkdownEditor value={value} onChange={onChange} />;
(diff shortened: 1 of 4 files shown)


src/editors/MarkdownEditor/index.tsx
```diff
@@ -2,12 +2,12 @@ import { FC } from 'react';
...
 type MarkdownInputProps = {
-  value: string;
+  value?: string;
   onChange: (newValue: string) => void;
 };
 
 const MarkdownInput: FC<MarkdownInputProps> = ({ value, onChange }) => {
-  return <MarkdownEditor value={value} onChange={onChange} />;
+  return <MarkdownEditor value={value ?? ''} onChange={onChange} />;
 };
```
mad-raccoon, line 11:
```
const MarkdownInput: FC<MarkdownInputProps> = ({ value = '', onChange }) => {
  return <MarkdownEditor value={value} onChange={onChange} />;
};
```

This removes the need for value ?? '' in JSX and keeps the fallback logic closer to the prop definition, improving readability.

In what cases will the value become undefined? I plan this component to receive a string always to avoid this kind of situations

mad-raccoon (changes requested):

PR #143, opened 2026-03-31
[RUI-245]: Add support for sort and limit for Bar charts

alicjakujawa:

Builders want to be able to set a sort order for the stacked and grouped bar charts based on the x-axis bar / group totals

# Main changes

Added "Sort by x-axis total" and "Limit x-axis items" inputs to stacked and grouped bar charts (vertical + horizontal).
When configured, a two-query flow runs: first query fetches aggregated category totals with backend sort/limit, then the main query fetches only the matching categories via a backend filter. Chart bars render in the totals query order.

# Test evidence

https://github.com/user-attachments/assets/7e49a012-6f77-4a08-9b9a-f6f7381b49f6

## Release Notes

* **New Features**
  * Bar chart components now support sorting by axis total (ascending or descending) to reorder categories based on their group totals.
  * Bar chart components now support limiting the number of displayed categories, allowing you to show only the top or bottom categories based on group totals.


src/components/component.inputs.constants.ts
```diff
@@ -416,6 +417,21 @@ const yAxisMaxItems = {
...
+  category: 'Component Settings',
+} as const;
+
+const limitAxisItems = {
+  name: 'limitAxisItems',
+  type: 'number',
+  label: 'Limit axis items',
+  description: 'Load only the top or bottom categories, based on group totals',
+  category: 'Component Settings',
+} as const;
```
mad-raccoon, line 448: i think this will be confusing for the builder users - Limit axis items doesn't specify x or y.

lets check with @hjkmarshall

src/components/component.inputs.constants.ts
```diff
@@ -416,6 +417,21 @@ const yAxisMaxItems = {
   category: 'Axes Settings',
 } as const;
 
+const sortByAxisTotal = {
```
mad-raccoon: this name and the label can be misleading. this dictates the sort direction instead of the sortBy

src/components/types/SortDirection.type.emb.ts
```diff
@@ -0,0 +1,11 @@
+import { defineOption, defineType } from '@embeddable.com/core';
+
+const SortDirectionType = defineType('sortDirection', {
+  label: 'Sort direction',
+  optionLabel: (value: string) => value,
+});
+
+defineOption(SortDirectionType, 'Ascending');
+defineOption(SortDirectionType, 'Descending');
```
mad-raccoon: if i am not mistaken you can already specify the real values and just change the optionLabel. something like this:


```suggestion
const SortDirectionType = defineType('sortDirection', {
  label: 'Sort direction',
  optionLabel: (value: string) => value === 'asc'? 'Ascending' : 'Descending',
});

defineOption(SortDirectionType, 'asc');
defineOption(SortDirectionType, 'desc');
```

this will avoid having to use the SORT_DIRECTION_MAP  or the parser function

src/components/charts/bars/bars.loadData.utils.ts
```diff
@@ -0,0 +1,77 @@
...
+};
+
+export const loadDataMainArgs = (
+  dataset: Dataset,
+  axis: Dimension,
+  groupBy: Dimension,
+  measure: Measure,
+  maxResults: number,
+  axisItems?: string[],
+): LoadDataRequest => {
```
mad-raccoon: same here for the args

src/components/charts/bars/bars.loadData.utils.ts
```diff
@@ -0,0 +1,77 @@
...
+  sortByAxisTotal?: string,
+  limitAxisItems?: number,
+): LoadDataRequest => {
+  const direction = parseSortDirection(sortByAxisTotal) ?? 'desc';
+
+  return {
+    from: dataset,
+    select: [axis, measure],
+    orderBy: [{ property: measure, direction }],
+    limit: isValidLimit(limitAxisItems) ? limitAxisItems : undefined,
```
mad-raccoon: we can have a getLimit function instead

src/components/charts/bars/bars.loadData.utils.ts
```diff
@@ -0,0 +1,77 @@
...
+  !!parseSortDirection(sortByAxisTotal) || isValidLimit(limitAxisItems);
+
+export const loadDataTotalsArgs = (
+  dataset: Dataset,
+  axis: Dimension,
+  measure: Measure,
+  sortByAxisTotal?: string,
+  limitAxisItems?: number,
+): LoadDataRequest => {
+  const direction = parseSortDirection(sortByAxisTotal) ?? 'desc';
```
mad-raccoon: i think this will no longer be needed

src/components/charts/bars/bars.loadData.utils.ts
```diff
@@ -0,0 +1,77 @@
...
+export const hasSortOrLimit = (sortByAxisTotal?: string, limitAxisItems?: number): boolean =>
+  !!parseSortDirection(sortByAxisTotal) || isValidLimit(limitAxisItems);
+
+export const loadDataTotalsArgs = (
+  dataset: Dataset,
+  axis: Dimension,
+  measure: Measure,
+  sortByAxisTotal?: string,
+  limitAxisItems?: number,
+): LoadDataRequest => {
```
mad-raccoon: for functions that receive more than 3 args, lets prefer objects, is better for clarity, flexibility and future

src/components/charts/bars/bars.loadData.utils.ts
```diff
@@ -0,0 +1,77 @@
...
+  OrderDirection,
+} from '@embeddable.com/core';
+
+const SORT_DIRECTION_MAP: Record<string, OrderDirection> = {
+  Ascending: 'asc',
+  Descending: 'desc',
+};
+
+export const parseSortDirection = (value?: string): OrderDirection | undefined =>
+  value ? SORT_DIRECTION_MAP[value] : undefined;
```
mad-raccoon: i believe this will no longer be needed

src/components/charts/bars/bars.hooks.ts
```diff
@@ -0,0 +1,20 @@
+import { useEffect } from 'react';
+import { DataResponse, Dimension } from '@embeddable.com/core';
+
+export function useSyncAxisItems(
+  resultsTotals: DataResponse | undefined,
+  axisDimension: Dimension,
+  setAxisItems: (values: string[], key: string) => void,
```
mad-raccoon: lets make this nullable, so we dont need to do this    **props.setAxisItems ?? (() => {}),**

and on the useEffect update this:

`   if (!setAxisItems || !resultsTotals?.data || resultsTotals.isLoading || !currentTotalsKey) return;`

also, consider passing an object instead of 4 args for the bellow reasons

and if there is a better name than useSyncAxisItems lets use it - e.g. useUpdateAxisOrder

(question, i see we setting the currentTotalsKey, but the state already has its value - do we need it here too ?)

src/components/charts/bars/BarChartStackedPro/index.tsx
```diff
@@ -67,6 +87,7 @@ const BarChartStackedPro = (props: BarChartStackedProProps) => {
       dimension: xAxis,
       groupDimension: groupBy,
       measure,
+      axisOrder: props.axisItems,
```
mad-raccoon: lets avoid the usage of props.axisItems, 

instead lets extract it here: **const { hideMenu , axisItems } = props;**

src/components/charts/bars/BarChartStackedPro/index.tsx
```diff
@@ -56,8 +61,23 @@ const BarChartStackedPro = (props: BarChartStackedProProps) => {
...
+    props.currentTotalsKey,
+  );
+
+  const resultsResponse =
+    props.results ??
+    ({
+      isLoading: !props.resultsTotals?.error,
+      data: [],
+      error: props.resultsTotals?.error,
+    } as DataResponse);
```
mad-raccoon: humm... why do we need  this ?

src/components/charts/bars/BarChartStackedPro/index.tsx
```diff
@@ -56,8 +61,23 @@ const BarChartStackedPro = (props: BarChartStackedProProps) => {
 
   const { hideMenu } = props;
 
+  useSyncAxisItems(
+    props.resultsTotals,
+    xAxis,
+    props.setAxisItems ?? (() => {}),
+    props.currentTotalsKey,
```
mad-raccoon: lets use the object destructuring 

`const { hideMenu, setAxisItems, currentTotalsKey } = props;`

src/components/charts/bars/BarChartStackedPro/index.tsx
```diff
@@ -56,8 +61,23 @@ const BarChartStackedPro = (props: BarChartStackedProProps) => {
 
   const { hideMenu } = props;
 
+  useSyncAxisItems(
+    props.resultsTotals,
+    xAxis,
+    props.setAxisItems ?? (() => {}),
```
mad-raccoon: we can avoid the coalescing, and make the hook deal with it

src/components/charts/bars/BarChartStackedPro/index.tsx
```diff
@@ -9,12 +9,17 @@ import { mergician } from 'mergician';
...
+import { useSyncAxisItems } from '../bars.hooks';
 
 export type BarChartStackedProProps = {
   groupBy: Dimension;
   maxLegendItems?: number;
   measure: Measure;
-  results: DataResponse;
+  results?: DataResponse;
+  resultsTotals?: DataResponse;
+  axisItems?: string[];
```
mad-raccoon: same here, i would specify something like axisOrder - axis items doesnt ring a bell

src/components/charts/bars/BarChartStackedPro/index.tsx
```diff
@@ -9,12 +9,17 @@ import { mergician } from 'mergician';
...
 import { ChartGranularitySelectField } from '../../shared/ChartGranularitySelectField/ChartGranularitySelectField';
+import { useSyncAxisItems } from '../bars.hooks';
 
 export type BarChartStackedProProps = {
   groupBy: Dimension;
   maxLegendItems?: number;
   measure: Measure;
-  results: DataResponse;
+  results?: DataResponse;
+  resultsTotals?: DataResponse;
```
mad-raccoon: i wonder if resultsAxisOrder would be a better/clearer name - totals doesnt make the connection to the gather of the top X items sorted by something.

src/components/charts/bars/BarChartStackedPro/definition.ts
```diff
@@ -91,12 +127,33 @@ const props = (
...
+    currentTotalsKey === state?.axisItemsKey &&
+    state?.axisItems != null;
 
   return {
     ...inputs,
     xAxis: xAxisWithGranularity,
-    setGranularity: (granularity: Granularity) => setState({ granularity }),
-    results: loadDataResults(inputs, xAxisWithGranularity),
+    setGranularity: (granularity: Granularity) => setState({ ...state, granularity }),
+    resultsTotals: totalsArgs ? loadData(totalsArgs) : undefined,
```
mad-raccoon: considering that this component implementation is quite tricky, i would simplify it like this:

```
const loadDataResultsTotal = () => {
    if(shouldGetTopItems) {
        return loadDataResultsTotalsArgs(inputs, xAxisWithGranularity);
    } 
   return undefined;
}

const loadDataResults = () => {
    if(shouldGetTopItems) {
        return loadDataResultsTotalsArgs(inputs, xAxisWithGranularity);
    } 
   return undefined;
}
...
resultsTotal: loadDataResultsTotal(...)
results: loadDataResults(...)
...
```

src/components/charts/bars/BarChartStackedPro/definition.ts
```diff
@@ -91,12 +127,33 @@ const props = (
...
+
+  const totalsArgs = needsSortLimit
+    ? loadDataResultsTotalsArgs(inputs, xAxisWithGranularity)
+    : undefined;
+  const currentTotalsKey = totalsArgs ? JSON.stringify(totalsArgs) : undefined;
+
+  const axisItemsFresh =
+    currentTotalsKey != null &&
+    currentTotalsKey === state?.axisItemsKey &&
+    state?.axisItems != null;
```
mad-raccoon: this could be done inside the loadDataResults

mad-raccoon: and be reused across components

src/components/charts/bars/BarChartStackedPro/definition.ts
```diff
@@ -91,12 +127,33 @@ const props = (
   [state, setState]: [BarChartStackedProState, (state: BarChartStackedProState) => void],
 ) => {
   const xAxisWithGranularity = getDimensionWithGranularity(inputs.xAxis, state?.granularity);
+  const needsSortLimit = hasSortOrLimit(
```
mad-raccoon: this function is to determine the flow of the loadData correct - if it fires 2 loadDatas or not correct ? if so lets rename is accordingly so it is more clear

e.g. const shouldGetTopItems = (): boolean => ...

src/components/charts/bars/BarChartStackedPro/definition.ts
```diff
@@ -70,14 +80,40 @@ const previewConfig = {
...
+  inputs: Inputs<typeof meta>,
+  xAxis: Dimension,
+  axisItems?: string[],
+): DataResponse => loadData(loadDataResultsArgs(inputs, xAxis, axisItems));
+
+const loadDataResultsTotalsArgs = (
+  inputs: Inputs<typeof meta>,
+  xAxis: Dimension,
+): LoadDataRequest =>
+  loadDataTotalsArgs(
```
mad-raccoon: consider the naming **buildLoadDataTotalsArgs** instead of **loadDataTotalsArgs**

mad-raccoon: i think we can even get rid of this middle function loadDataResultsTotalsArgs

and only have the **loadDataTotalsArgs** renamed to **loadDataResultsTotalsArgs**

src/components/charts/bars/BarChartStackedPro/definition.ts
```diff
@@ -70,14 +80,40 @@ const previewConfig = {
...
+  xAxis?: Dimension,
+  axisItems?: string[],
+): LoadDataRequest =>
+  loadDataMainArgs(
+    inputs.dataset,
+    xAxis ?? inputs.xAxis,
+    inputs.groupBy,
+    inputs.measure,
+    inputs.maxResults,
+    axisItems,
```
mad-raccoon: lets use object instead of multiple args

consider the naming **buildLoadDataResultsArgs** instead of **loadDataMainArgs**


i think we can even get rid of this middle function loadDataResultsArgs

and only have the loadDataMainArgs renamed to loadDataResultsArgs

src/components/charts/bars/BarChartStackedPro/definition.ts
```diff
@@ -58,6 +66,8 @@ const meta = {
 
 export type BarChartStackedProState = {
   granularity?: Granularity;
+  axisItems?: string[];
+  axisItemsKey?: string;
```
mad-raccoon: i think we can rename this to something more clear

mad-raccoon (changes requested):
hey! I reviewed from bottom to top and focused on the BarChartStacked component only, assuming the same approach applies to the others.

The goal is to simplify the implementation and make it more readable/understandable.

There’s quite a bit in there, so feel free to flag anything that’s unclear. Happy to walk through it if helpful. I also might have miss something.

src/components/types/SortDirection.type.emb.ts
alicjakujawa: so when i implemented it, it broke the dropdown (i wasnt able to select Descending option)

src/components/component.inputs.constants.ts
```diff
@@ -416,6 +417,21 @@ const yAxisMaxItems = {
...
 } as const;
 
+const sortDirectionTopAxis = {
+  name: 'sortDirectionTopAxis',
+  type: SortDirectionType,
+  label: 'Sort direction top axis',
+  category: 'Component Settings',
+} as const;
+
+const limitAxisItems = {
```
mad-raccoon: to match the other one: **limitTopAxis**

src/components/charts/bars/bars.loadData.utils.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (value === 'Ascending') return 'asc';
+  if (value === 'Descending') return 'desc';
+  return undefined;
+};
+
+export const getLimit = (limit?: number): number | undefined =>
+  typeof limit === 'number' && Number.isInteger(limit) && limit > 0 ? limit : undefined;
+
+export const shouldGetTopItems = (sortDirection?: string, limit?: number): boolean =>
+  toOrderDirection(sortDirection) != null || getLimit(limit) != null;
```
mad-raccoon: i think this can be simplified:

```suggestion
  toOrderDirection(sortDirection) || getLimit(limit);
```

src/components/charts/bars/bars.hooks.ts
```diff
@@ -0,0 +1,28 @@
+import { useEffect } from 'react';
+import { DataResponse, Dimension } from '@embeddable.com/core';
+
+export function useUpdateAxisOrder(opts: {
+  resultsAxisOrder?: DataResponse;
+  axisDimension: Dimension;
+  setAxisOrder?: (values: string[], key: string) => void;
+  currentAxisOrderKey?: string;
```
mad-raccoon: we already have this in the state, why do we need to update the state with the same value ?

 **setAxisOrder(values, currentAxisOrderKey);**

src/components/component.inputs.constants.ts
alicjakujawa: We agreed with @hjkmarshall to update it to  Limit x-axis items

src/components/charts/bars/BarChartGroupedHorizontalPro/definition.ts
```diff
@@ -93,12 +129,37 @@ const props = (
...
+        dataset: inputs.dataset,
+        axis: yAxisWithGranularity,
+        measure: inputs.measure,
+        sortDirection,
+        limit: inputs.limitAxisItems,
+      })
+    : undefined;
+  const currentAxisOrderKey = axisOrderArgs ? JSON.stringify(axisOrderArgs) : undefined;
+
+  const axisOrderFresh =
```
mad-raccoon: this code is duplicated

src/components/charts/bars/BarChartGroupedHorizontalPro/definition.ts
```diff
@@ -93,12 +129,37 @@ const props = (
   ],
 ) => {
   const yAxisWithGranularity = getDimensionWithGranularity(inputs.yAxis, state?.granularity);
+  const sortDirection = inputs.sortDirectionTopAxis as string | undefined;
+  const needsTopItems = shouldGetTopItems(sortDirection, inputs.limitAxisItems);
+
+  const axisOrderArgs = needsTopItems
```
mad-raccoon: this code is duplicated -  i believe we can create reusable function for this

src/components/charts/bars/BarChartGroupedHorizontalPro/definition.ts
```diff
@@ -69,14 +79,40 @@ const previewConfig = {
...
+  inputs: Inputs<typeof meta>,
+  yAxis: Dimension,
+): LoadDataRequest =>
+  buildAxisOrderArgs({
+    dataset: inputs.dataset,
+    axis: yAxis,
+    measure: inputs.measure,
+    sortDirection: inputs.sortDirectionTopAxis as string | undefined,
+    limit: inputs.limitAxisItems,
+  });
```
mad-raccoon: lets just rename the buildAxisOrderArgs to loadDataResultsAxisOrderArgs, and reuse it across components, wdyt ? 

otherwise we will have a lot of duplicated code that doesnt seem usefull

src/components/component.inputs.constants.ts
mad-raccoon: we need to create this two objects then - or we create a reusable one, and we just change the name and label

src/components/charts/bars/bars.hooks.ts
```diff
@@ -0,0 +1,28 @@
+import { useEffect } from 'react';
+import { DataResponse, Dimension } from '@embeddable.com/core';
+
+export function useUpdateAxisOrder(opts: {
+  resultsAxisOrder?: DataResponse;
+  axisDimension: Dimension;
+  setAxisOrder?: (values: string[], cacheKey: string) => void;
```
mad-raccoon: we can update this name now to be setAxisOrderAndCacheKey, so it matches the rest

src/components/charts/bars/BarChartGroupedHorizontalPro/index.tsx
```diff
@@ -53,10 +58,29 @@ const BarChartGroupedHorizontalPro = (props: BarChartGroupedHorizontalProProps)
...
+    axisOrder,
+    resultsAxisOrder,
+    axisOrderCacheKey,
+    setAxisOrderAndCacheKey,
   } = props;
 
+  useUpdateAxisOrder({
+    resultsAxisOrder,
+    axisDimension: yAxis,
+    setAxisOrder: setAxisOrderAndCacheKey,
```
mad-raccoon:
```suggestion
    setAxisOrderAndCacheKey
```

src/components/charts/bars/BarChartStackedPro/index.tsx
mad-raccoon: why do we need this *??* 
from what i see or we are collecting data or not - trying to understand why this is needed

src/components/charts/bars/bars.hooks.ts
```diff
@@ -0,0 +1,28 @@
+import { useEffect } from 'react';
+import { DataResponse, Dimension } from '@embeddable.com/core';
+
+export function useUpdateAxisOrder(opts: {
```
mad-raccoon: useUpdateAxisOrderAndCacheKey - so it matches

mad-raccoon (commented):
only a few comments lets - this is more cleanup and naming - it looks good to me 🚀

embeddable.config.ts
mad-raccoon, line 1: we need to rollback this file

mad-raccoon (changes requested):

src/components/charts/bars/BarChartStackedPro/index.tsx
alicjakujawa: this is guard to not pass undefine to useFillGaps hook, but we could move this fallback to hook itself

mad-raccoon: the useFillGaps already includes this fallback

mad-raccoon (approved):
🔥

PR #182, opened 2026-05-07
[RUI-269]: Add description field to all definition files

alicjakujawa:

All the RPRO component should now contain a description on what the chart is about. We need this for dashboard-as-code  feature.

# Main changes

Added description field to all components in RPRO

## Release Notes

* **Documentation**
  * Added descriptive metadata across all chart components (bar, line, pie, scatter, heat map, table, and more), providing clear guidance on use cases and data requirements.
  * Added descriptions to all editor and input components, clarifying their purpose and variable pairings.
  * Enhanced shared utility components with helpful descriptions for layout and formatting.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@embeddable.com/remarkable-pro",
-  "version": "0.2.10",
+  "version": "0.2.13",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@embeddable.com/remarkable-pro",
-      "version": "0.2.10",
+      "version": "0.2.13",
       "dependencies": {
         "@changesets/cli": "^2.29.6",
         "@embeddable.com/core": "^2.13.7",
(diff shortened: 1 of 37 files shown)


mikhail-terekhov (changes requested):
@alicjakujawa could you pls bump the sdk versions? we added this description field recently.
@mad-raccoon also I can see that sdk-core and sdk-react versions are fixed. Is there any reason for that? Let’s change this to `^` unless there’s a specific reason not to. If there is, please let me know

mikhail-terekhov (approved):

mikhail-terekhov (commented):
comment from Harry: 
@Mikhail these are ok descriptions for when to use components, but they won't help the AI know how to use the components.

There are edge-cases that I mentioned above:


e.g. comparison component datasets shouldn't be filtered by timerange variables - there are component-specific inputs to receive those variables instead. 
e.g. when the comparison line chart components take a non-time-based dimension, you need to specify via another input which time dimension should be used for the comparison.  
e.g. if you add a title to a control component, it needs more height. 


Any thoughts on how to deal with these types of things?

Tom and I were discussing it. We thought a good approach would be for descriptions to help the AI choose when to use each component. And then to have separate "how to use" context which it can pull it based on the ones it chooses.

-----
I will think about this today. @alicjakujawa pls let me know if you have some ideas
For the context, the thread is here https://team-embeddable.slack.com/archives/C0B0GC984UV/p1778514458127889?thread_ts=1778079154.484709&cid=C0B0GC984UV

mikhail-terekhov (approved):
Everithing looks and works goood. Let's release that change. Could you please add a changeset file with a minor version bump and release it?

PR #184, opened 2026-05-11
[RUI-240]: Add BarLineChartPro component

alicjakujawa:
We need chart type that supports multiple bar measures and multiple line measures plotted against a single shared dimension, with optional secondary Y-axis for line series.

# Main changes

Added BarLineChartPro component

# Test evidence

https://github.com/user-attachments/assets/700be596-e7b9-4a48-b8ab-0c0091ebd559

## Summary by CodeRabbit

* **New Features**
  * Added Bar + Line Chart component for visualizing multiple measures as bars and lines simultaneously on the same chart.
  * Support for secondary y-axis with customizable labels and ranges.
  * Customizable line styling including dashed lines and fill-under-line options.
  * Click event handlers for bar and line interactions.

### DIFF
--- a/src/components/charts/combo/BarLineChartPro/BarLineChartPro.utils.test.ts
+++ b/src/components/charts/combo/BarLineChartPro/BarLineChartPro.utils.test.ts
@@ -1,0 +1,878 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import type { Dimension, Measure } from '@embeddable.com/core';
+import {
+  getBarLineChartProData,
+  getBarLineChartProOptions,
+  createBarLineClickHandler,
+} from './BarLineChartPro.utils';
+import { getThemeFormatter } from '../../../../theme/formatter/formatter.utils';
+import { getDimensionWithoutTruncation, groupTailAsOther } from '../../charts.utils';
+import { getDimensionMeasureColor } from '../../../../theme/styles/styles.utils';
+import { getTimeRangeFromDimensionValue } from '../../../utils/dimension.utils';
+import { getBarChartProOptions } from '../../bars/bars.utils';
+
+vi.mock('../../../../theme/formatter/formatter.utils', () => ({ getThemeFormatter: vi.fn() }));
+vi.mock('../../charts.utils', () => ({
+  groupTailAsOther: vi.fn(),
+  getDimensionWithoutTruncation: vi.fn((d) => d),
+}));
+vi.mock('../../../../theme/styles/styles.utils', () => ({ getDimensionMeasureColor: vi.fn() }));
+vi.mock('@embeddable.com/remarkable-ui', () => ({
+  getChartColors: vi.fn(() => []),
+  getStyleNumber: vi.fn(() => 4),
+}));
+vi.mock('../../../../utils/color.utils', () => ({
+  isColorValid: vi.fn(() => false),
+  setColorAlpha: vi.fn((color: string) => `${color}-alpha`),
+}));
+vi.mock('../../bars/bars.utils', () => ({ getBarChartProOptions: vi.fn(() => ({})) }));
+vi.mock('../../../utils/dimension.utils', () => ({ getTimeRangeFromDimensionValue: vi.fn() }));
+vi.mock('mergician', () => ({
+  mergician: vi.fn((...args: object[]) => Object.assign({}, ...args)),
+}));
+
+// -- helpers -----------------------------------------------------------------
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const makeDimension = (overrides: Record<string, any> = {}): Dimension =>
+  ({
+    name: 'date',
+    title: 'Date',
+    nativeType: 'string',
+    inputs: {},
+    ...overrides,
+  }) as unknown as Dimension;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const makeMeasure = (overrides: Record<string, any> = {}): Measure =>
+  ({
+    name: 'revenue',
+    title: 'Revenue',
+    nativeType: 'number',
+    inputs: {},
+    ...overrides,
+  }) as unknown as Measure;
+
+const makeTheme = (overrides = {}) => ({ charts: {}, ...overrides }) as never;
+
+const makeMockFormatter = () => ({
+  data: vi.fn((_: unknown, value: unknown) => `fmt:${value}`),
+  dimensionOrMeasureTitle: vi.fn((m: { title: string }) => m.title),
+});
+
+// ----------------------------------------------------------------------------
+
+describe('getBarLineChartProData', () => {
+  let mockFormatter: ReturnType<typeof makeMockFormatter>;
+
+  beforeEach(() => {
+    mockFormatter = makeMockFormatter();
+    vi.mocked(getThemeFormatter).mockReturnValue(mockFormatter as never);
+    vi.mocked(getDimensionMeasureColor).mockImplementation(({ color, index }) =>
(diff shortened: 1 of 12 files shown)


src/components/charts/combo/BarLineChartPro/BarLineChartPro.utils.ts
```diff
@@ -0,0 +1,263 @@
...
+      theme,
+      color: 'border',
+      value: measure.name,
+      index,
+      chartColors,
+    }),
+  }));
+
+  const lineDatasets = lineMeasures.map((measure, index) => {
+    const colorIndex = measures.length + index;
```
mad-raccoon: why do we need this ? shouldn't we use the index as it is ?

src/components/charts/combo/BarLineChartPro/BarLineChartPro.utils.ts
```diff
@@ -0,0 +1,263 @@
...
+  return {
+    labels,
+    datasets: [...barDatasets, ...lineDatasets] as ChartData<'bar'>['datasets'],
+  };
+};
+
+export const getBarLineChartProOptions = (
+  options: {
+    measures: Measure[];
+    lineMeasures: Measure[];
```
mad-raccoon: could we make is more explicit:

barMeasures
lineMeasures

src/components/charts/combo/BarLineChartPro/BarLineChartPro.utils.ts
```diff
@@ -0,0 +1,263 @@
...
+    data,
+    showSecondaryAxis,
+    showValueLabels,
+    showValueLabelsLine,
+    yAxisSecondaryLabel,
+    yAxisSecondaryMin,
+    yAxisSecondaryMax,
+  } = options;
+
+  const allMeasures = [...measures, ...lineMeasures];
```
mad-raccoon:
```suggestion
  const measures = [...barMeasures, ...lineMeasures];
```

src/components/charts/combo/BarLineChartPro/BarLineChartPro.utils.ts
```diff
@@ -0,0 +1,263 @@
...
+    },
+  };
+
+  const tooltipOptions: Partial<ChartOptions<'bar'>> = {
+    plugins: {
+      tooltip: {
+        callbacks: {
+          title: (context) => {
+            const label = context[0]?.label;
+            return themeFormatter.data(getDimensionWithoutTruncation(dimension), label);
```
mad-raccoon, line 182: 🔥

src/components/charts/combo/BarLineChartPro/definition.ts
```diff
@@ -0,0 +1,177 @@
...
+      name: 'lineMeasures',
+      label: 'Line measures',
+      required: false,
+      inputs: [
+        ...inputs.measures.inputs,
+        { ...subInputs.color, name: 'lineColor', label: 'Line color' },
+        { ...subInputs.boolean, name: 'dashedLine', label: 'Dashed line', defaultValue: false },
+        { ...subInputs.boolean, name: 'connectGaps', label: 'Connect gaps', defaultValue: true },
+        { ...subInputs.boolean, name: 'fillUnderLine', label: 'Fill under line' },
+        { ...subInputs.boolean, name: 'useSecondaryAxis', label: 'Use secondary axis' },
```
mad-raccoon: out of curiosity, why do we need this useSecondaryAxis ?

src/components/charts/combo/BarLineChartPro/definition.ts
```diff
@@ -0,0 +1,177 @@
...
+import { getDimensionWithGranularity } from '../../utils/granularity.utils';
+import { getClientContextTimezone } from '../../../../theme/utils/clientContext.utils';
+import { ThemeClientContext } from '../../../../theme/theme.types';
+import { BarLineChartProClickArg } from '../combo.types';
+
+const meta = {
+  name: 'BarLineChartPro',
+  label: 'Bar + Line Chart',
+  category: 'Combo Charts',
+  inputs: [
```
mad-raccoon, line 23: would it be possible to reuse the inputs of the barChartDefaultPro.meta and lineChartDefaultPro.meta

something identically to what we did with the "Line chart with kpis" - this way you would not have to define as much

mad-raccoon (changes requested):
looking good !

src/components/charts/combo/BarLineChartPro/BarLineChartPro.utils.ts
alicjakujawa: its to have different colors for bars/lines

src/components/charts/combo/BarLineChartPro/definition.ts
alicjakujawa: in task req there is "Optional secondary Y-axis"

src/components/charts/combo/BarLineChartPro/definition.ts
alicjakujawa: We could but it would be a bit of IFology because of it, something like:

  inputs: [
    ...barChartDefaultPro.meta.inputs.flatMap((input) => {
      if (input.name === 'measures') {
        return [
          { ...input, label: 'Bar measures' },
          lineMeasuresInput,
        ];
      }
      if (input.name === 'showValueLabels') {
        return [input, showValueLabelsLineInput];
      }
      if (input.name === 'maxResults') {
        return [...secondaryAxisInputs, input];
      }
      return [input];
    }),
  ],

src/components/charts/combo/BarLineChartPro/BarLineChartPro.utils.test.ts
```diff
@@ -0,0 +1,878 @@
...
+}));
+vi.mock('../../../../utils/color.utils', () => ({
+  isColorValid: vi.fn(() => false),
+  setColorAlpha: vi.fn((color: string) => `${color}-alpha`),
+}));
+vi.mock('../../bars/bars.utils', () => ({ getBarChartProOptions: vi.fn(() => ({})) }));
+vi.mock('../../../utils/dimension.utils', () => ({ getTimeRangeFromDimensionValue: vi.fn() }));
+vi.mock('mergician', () => ({
+  mergician: vi.fn((...args: object[]) => Object.assign({}, ...args)),
+}));
```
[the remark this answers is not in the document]
mad-raccoon: can we simply use mergician ?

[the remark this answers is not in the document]
mad-raccoon: and not mock it

src/components/charts/combo/BarLineChartPro/definition.ts
mad-raccoon: you can do differently - 1st check to which one it is more identical to bars or lines - and use that one as a base

imagine it is bars

```
inputs: [
   ...barchart.meta,
   linechart.meta.find(x=>x.name === 'dimensions)
   ...
] 
```

let me know if this makes sense.

the idea here was to reuse suff from both line and bar charts, since they are already defined (inputs ans subinputs)

alicjakujawa: but they have different order and some different inputs or labels, so we are not able to do pure reuse, we need to "update" some of them thats why those IFs. Which does not feel right -> we was talking about meta being something for customers to see and need to understand

mad-raccoon: okay !

mad-raccoon (approved):

PR #208, opened 2026-05-29
[RUI-280]: Add export options to component settings in builder

alicjakujawa:
We need to allow users to enable / disable export options for each chart.

# Main changes
Added new input exportOptions
Updated ChartCardMenuPro to properly display options based on what was selected

# Test evidence

https://github.com/user-attachments/assets/9285e25a-29d0-4dd9-b04a-d2c41f083f2a

## Summary by CodeRabbit

* **New Features**
  * Added export options configuration to chart components, enabling users to customize which export formats (CSV, XLSX, PNG) are available in the chart export menu. Export options can now be selected and configured in the component builder settings.

### DIFF
--- a/src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
+++ b/src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
@@ -1,7 +1,7 @@
 import { useTheme } from '@embeddable.com/react';
 import { Theme } from '../../../../theme/theme.types';
 import { i18nSetup } from '../../../../theme/i18n/i18n';
-import { ChartCard, pickChartCardHeaderProps } from '../../shared/ChartCard/ChartCard';
+import { ChartCard, asChartCardHeaderProps } from '../../shared/ChartCard/ChartCard';
 import { resolveI18nProps } from '../../../component.utils';
 import { BarChart } from '@embeddable.com/remarkable-ui';
 import { getBarChartProData, getBarChartProOptions } from '../bars.utils';
@@ -38,8 +38,8 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps) 
     onBarClicked,
   } = props;
 
-  const resolvedProps = resolveI18nProps(props);
-  const { description, title, tooltip, xAxisLabel, yAxisLabel } = resolvedProps;
+  const resolvedI18nProps = resolveI18nProps(props);
+  const { description, title, tooltip, xAxisLabel, yAxisLabel } = resolvedI18nProps;
 
   const results = useFillGaps({
     results: props.results,
@@ -70,7 +70,7 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps) 
       data={results}
       dimensionsAndMeasures={[dimension, ...measures]}
       errorMessage={results.error}
-      {...pickChartCardHeaderProps(resolvedProps)}
+      {...asChartCardHeaderProps(resolvedI18nProps)}
     >
       {setGranularity && (
         <ChartGranularitySelectField
--- a/src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
+++ b/src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
@@ -24,6 +24,7 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps) 
 
   const {
     hideMenu,
+    exportOptions,
     dimension,
     measures,
     showValueLabels,
@@ -74,6 +75,7 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps) 
       title={title}
       tooltip={tooltip}
       hideMenu={hideMenu}
+      exportOptions={exportOptions}
     >
       {setGranularity && (
         <ChartGranularitySelectField
--- a/src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
+++ b/src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
@@ -70,7 +70,7 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps) 
       data={results}
       dimensionsAndMeasures={[dimension, ...measures]}
       errorMessage={results.error}
-      {...asChartCardHeaderProps(resolvedI18nProps)}
+      {...asChartCardHeaderProps(resolvedI18nProps, props)}
     >
       {setGranularity && (
         <ChartGranularitySelectField
--- a/src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
+++ b/src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
@@ -1,7 +1,7 @@
 import { useTheme } from '@embeddable.com/react';
 import { Theme } from '../../../../theme/theme.types';
 import { i18nSetup } from '../../../../theme/i18n/i18n';
-import { ChartCard } from '../../shared/ChartCard/ChartCard';
+import { ChartCard, pickChartCardHeaderProps } from '../../shared/ChartCard/ChartCard';
 import { resolveI18nProps } from '../../../component.utils';
 import { BarChart } from '@embeddable.com/remarkable-ui';
 import { getBarChartProData, getBarChartProOptions } from '../bars.utils';
@@ -23,8 +23,6 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps) 
   i18nSetup(theme);
 
   const {
-    hideMenu,
-    exportOptions,
     dimension,
     measures,
     showValueLabels,
@@ -40,7 +38,8 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps) 
     onBarClicked,
   } = props;
 
-  const { description, title, tooltip, xAxisLabel, yAxisLabel } = resolveI18nProps(props);
+  const resolvedProps = resolveI18nProps(props);
+  const { description, title, tooltip, xAxisLabel, yAxisLabel } = resolvedProps;
 
   const results = useFillGaps({
     results: props.results,
@@ -71,11 +70,7 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps) 
       data={results}
       dimensionsAndMeasures={[dimension, ...measures]}
       errorMessage={results.error}
-      description={description}
-      title={title}
-      tooltip={tooltip}
-      hideMenu={hideMenu}
-      exportOptions={exportOptions}
+      {...pickChartCardHeaderProps(resolvedProps)}
     >
       {setGranularity && (
         <ChartGranularitySelectField
--- a/src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
+++ b/src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
@@ -70,7 +70,7 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps) 
       data={results}
       dimensionsAndMeasures={[dimension, ...measures]}
       errorMessage={results.error}
-      {...asChartCardHeaderProps(resolvedI18nProps, props)}
+      {...asChartCardHeaderProps(props)}
     >
       {setGranularity && (
         <ChartGranularitySelectField
--- a/src/components/charts/bars/BarChartDefaultPro/index.tsx
+++ b/src/components/charts/bars/BarChartDefaultPro/index.tsx
@@ -39,8 +39,7 @@ const BarChartDefaultPro = (props: BarChartDefaultProProps) => {
   } = props;
 
   const resolvedI18nProps = resolveI18nProps(props);
-  const { xAxisLabel, yAxisLabel } = resolvedI18nProps;
-  const { tooltip, description, title } = resolvedI18nProps;
+  const { tooltip, description, title, xAxisLabel, yAxisLabel } = resolvedI18nProps;
 
   const results = useFillGaps({
     results: props.results,
@@ -71,7 +70,7 @@ const BarChartDefaultPro = (props: BarChartDefaultProProps) => {
       data={results}
       dimensionsAndMeasures={[dimension, ...measures]}
       errorMessage={results.error}
-      {...asChartCardHeaderProps(resolvedI18nProps)}
+      {...asChartCardHeaderProps(resolvedI18nProps, props)}
     >
       {setGranularity && (
         <ChartGranularitySelectField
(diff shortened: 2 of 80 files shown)


src/theme/defaults/defaults.ChartCardMenu.constants.ts
```diff
@@ -14,23 +14,27 @@ export type ChartCardMenuOptionOnClickProps = {
...
 export type ChartCardMenuOption = {
+  key: string;
   labelKey: string;
   iconSrc?: string;
   onClick: (props: ChartCardMenuOptionOnClickProps) => void;
 };
 
 export const defaultChartMenuProOptions: ChartCardMenuOption[] = [
   {
+    key: 'csv',
```
mad-raccoon: lets follow the same format as the other defaults, and use value instead of key

src/components/component.inputs.constants.ts
```diff
@@ -492,6 +492,22 @@ const yMeasure = {
...
+  label: 'Export options',
+  array: true,
+  defaultValue: ['csv', 'xlsx', 'png'],
+  category: 'Component Settings',
+  config: {
+    options: [
+      { value: 'csv', label: 'Export CSV' },
+      { value: 'xlsx', label: 'Export XLSX' },
+      { value: 'png', label: 'Export PNG' },
+    ],
```
mad-raccoon: we should avoid hardcoded values. imagine that lates we change a key/value - this will not raise any problem


lets follow the same approach as the granularity pls

mad-raccoon (changes requested):
a couple of comments, overall looks good.

only 2 points:

1. i would check with harry that this approach (of not having a type) is OK
2. since we are touching multiple files, i would take this opportunity, to make the **ChartCard** receive {...props} - this will clean up all the components - whats your take on this ?

src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
```diff
@@ -70,10 +70,7 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps)
       data={results}
       dimensionsAndMeasures={[dimension, ...measures]}
       errorMessage={results.error}
-      description={description}
-      title={title}
-      tooltip={tooltip}
-      hideMenu={hideMenu}
+      {...pickChartCardHeaderProps(resolvedProps)}
```
mad-raccoon: I'd lean towards asChartCardHeaderProps - the as prefix signals "extract/cast to this shape" which is exactly what it does.

src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
```diff
@@ -39,7 +38,8 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps)
     onBarClicked,
   } = props;
 
-  const { description, title, tooltip, xAxisLabel, yAxisLabel } = resolveI18nProps(props);
+  const resolvedProps = resolveI18nProps(props);
```
mad-raccoon: resolvedi18nProps > i would be more explicit - resolved is very generic

mad-raccoon (changes requested):

syedad218 (approved):

src/components/charts/shared/ChartCard/ChartCard.tsx
```diff
@@ -22,6 +22,14 @@ export type ChartCardHeaderProps = {
   description?: string;
   tooltip?: string;
   hideMenu?: boolean;
+  exportOptions?: (string | unknown)[];
```
ako-hk:
```suggestion
  exportOptions?: ('csv' | 'xlsx' | 'png')[];
```

src/components/charts/shared/ChartCard/ChartCard.tsx
```diff
@@ -103,6 +112,7 @@ export const ChartCard = React.forwardRef<HTMLDivElement, ChartCardProps>(
                   data={data?.data}
                   dimensionsAndMeasures={dimensionsAndMeasures}
                   onCustomDownload={onCustomDownload}
+                  enabledExportOptions={exportOptions as string[]}
```
ako-hk: neat:
```suggestion
                  exportOptions={exportOptions as string[]}
```
just to keep everything consistent across the repo.

src/components/charts/shared/ChartCard/ChartCardMenuPro/ChartCardMenuPro.tsx
```diff
@@ -24,15 +24,21 @@ export function InlineSvgFromData({ src, className, ...rest }: InlineSvgFromData
...
   i18nSetup(theme);
 
   const [isLoading, setIsLoading] = React.useState(false);
 
-  const options = theme.defaults.chartMenuOptions ?? [];
+  const allOptions = theme.defaults.chartMenuOptions ?? [];
+  const { enabledExportOptions } = props;
+  const options = enabledExportOptions
+    ? allOptions.filter((option) => enabledExportOptions.includes(option.value))
+    : allOptions;
```
ako-hk: This works today since chartMenuOptions only contains export items, but if a non-export action is ever added to the menu (e.g. Refresh, Share) it'll be silently hidden whenever enabledExportOptions is set. Suggest filtering only known export keys and leaving others through:

```suggestion
const EXPORT_VALUES = new Set(['csv', 'xlsx', 'png']);
const options = enabledExportOptions
  ? allOptions.filter(o => !EXPORT_VALUES.has(o.value) || enabledExportOptions.includes(o.value))
  : allOptions;
```

ako-hk (commented):
left some comments

ako-hk:
`AreaChartPro` is missing a `definition.ts` update. Without `inputs.exportOptions` in the definition, the setting won't appear in the builder — `exportOptions` will always be undefined here, so the feature silently does nothing for this chart. Same issue for `LineChartComparisonWithKpiTabsPro` and `LineChartWithKpiTabsPro`.

ako-hk:
>  [7.7% Duplication on New Code](https://sonarcloud.io/component_measures?id=embeddable-hq_remarkable-pro&pullRequest=208&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)

@alicjakujawa could you also reduce the duplicates?

src/components/charts/shared/ChartCard/ChartCard.tsx
alicjakujawa: @ako-hk So we are able to extend those types to whatever we want from boilerplate (json etc)

alicjakujawa:
> definition.ts

We are not able to reduce duplicates because we are touching all files with the same line of code :P

src/components/charts/shared/ChartCard/ChartCardMenuPro/ChartCardMenuPro.tsx
alicjakujawa: Hmmm again here, those options can be extended by customer from boilerplate.. Now when you are saying it im thinking if we should be so "fixed" to export options. Maybe it should be menuItem. And then we let customer to decide which to show which to hide. We wouldnt have this export / non export distinction. What you think?

ako-hk: Good point, that could be done

alicjakujawa:
> `AreaChartPro` is missing a `definition.ts` update. Without `inputs.exportOptions` in the definition, the setting won't appear in the builder — `exportOptions` will always be undefined here, so the feature silently does nothing for this chart. Same issue for `LineChartComparisonWithKpiTabsPro` and `LineChartWithKpiTabsPro`.

Those components extends inputs from base components, so export options included

ako-hk (approved):

src/components/charts/bars/BarChartDefaultPro/index.tsx
```diff
@@ -32,14 +32,15 @@ const BarChartDefaultPro = (props: BarChartDefaultProProps) => {
...
     dimension,
     granularity,
     setGranularity,
     onBarClicked,
   } = props;
 
-  const { tooltip, description, title, xAxisLabel, yAxisLabel } = resolveI18nProps(props);
+  const resolvedI18nProps = resolveI18nProps(props);
+  const { xAxisLabel, yAxisLabel } = resolvedI18nProps;
+  const { tooltip, description, title } = resolvedI18nProps;
```
mad-raccoon: this can be done in a single line

src/components/charts/shared/ChartCard/ChartCardMenuPro/ChartCardMenuPro.tsx
```diff
@@ -24,15 +24,21 @@ export function InlineSvgFromData({ src, className, ...rest }: InlineSvgFromData
...
 
 export const ChartCardMenuPro: React.FC<ChartCardMenuProProps> = (props) => {
   const theme: Theme = useTheme() as Theme;
   i18nSetup(theme);
 
   const [isLoading, setIsLoading] = React.useState(false);
 
-  const options = theme.defaults.chartMenuOptions ?? [];
+  const allOptions = theme.defaults.chartMenuOptions ?? [];
+  const { menuOptions } = props;
```
mad-raccoon, line 38: the only thing i dont like tis that the **menuOptions** as coming from the **const resolvedI18nProps = resolveI18nProps(props);**

and it is not really a resolved prop


do you have any ideas how we could improve this easily ?

src/components/charts/shared/ChartCard/ChartCard.tsx
```diff
@@ -22,6 +22,14 @@ export type ChartCardHeaderProps = {
   description?: string;
   tooltip?: string;
   hideMenu?: boolean;
+  menuOptions?: (string | unknown)[];
```
mad-raccoon, line 26: why unknown ?

src/components/charts/shared/ChartCard/ChartCard.tsx
```diff
@@ -103,6 +112,7 @@ export const ChartCard = React.forwardRef<HTMLDivElement, ChartCardProps>(
                   data={data?.data}
                   dimensionsAndMeasures={dimensionsAndMeasures}
                   onCustomDownload={onCustomDownload}
+                  menuOptions={menuOptions as string[]}
```
mad-raccoon: if we sort the types we will not need the alias

src/components/charts/tables/TableScrollable/definition.ts
```diff
@@ -61,6 +61,7 @@ const meta = {
       category: 'Component Settings',
     },
     { ...inputs.sortDirection, label: 'Default sort direction', category: 'Component Settings' },
+    { ...inputs.menuOptions, defaultValue: ['csv', 'xlsx'] },
```
mad-raccoon: instead of hardcoded ['csv', 'xlsx'], can we reuse some consts?

mad-raccoon: **ExportOption**

src/components/component.inputs.constants.ts
```diff
@@ -492,6 +493,15 @@ const yMeasure = {
   label: 'Y-axis measure',
 } as const;
 
+const menuOptions = {
+  name: 'menuOptions',
+  type: ExportOptionType,
+  label: 'Menu options',
+  array: true,
+  defaultValue: ['csv', 'xlsx', 'png'],
```
mad-raccoon: **ExportOption**

src/components/types/ExportOption.type.emb.ts
```diff
@@ -0,0 +1,18 @@
+import { defineOption, defineType } from '@embeddable.com/core';
+
+export const ExportOption = {
```
mad-raccoon: could we follow the pattern of the SortDirectionTypeOptions and DisplayFormatTypeOptions:

**ExportOptionTypeOptions**

mad-raccoon (changes requested):

src/components/charts/shared/ChartCard/ChartCard.tsx
alicjakujawa: because sdk is type it as unknown, as its customer types

src/components/charts/shared/ChartCard/ChartCard.tsx
alicjakujawa: commented above

src/components/charts/shared/ChartCard/ChartCardMenuPro/ChartCardMenuPro.tsx
mad-raccoon: not a fan of this approach, receiving both props and resolved props

i would simply receive props (ignore resolved)

src/components/charts/shared/ChartCard/ChartCard.tsx
mad-raccoon: you can make the asChartCardHeaderProps return a **string[]** , we control it. 

this sorts the bellow

src/components/charts/shared/ChartCard/ChartCardMenuPro/ChartCardMenuPro.tsx
mad-raccoon: (i think i suggested this at the beggining) there is no need to resolve title/subtitle/tooltip at the component level - instead you can make the ChartCard resolve it itself (after all, this is the guy who needs it resolved)

src/components/charts/shared/ChartCard/ChartCard.tsx
alicjakujawa: but we are not able to get rid of this unkown

src/components/charts/shared/ChartCard/ChartCard.tsx
alicjakujawa: updated here

src/components/charts/shared/ChartCard/ChartCard.tsx
mad-raccoon: i think we do, because now the menuOptions will come from the **asChartCardHeaderProps**. 

and on the **asChartCardHeaderProps** you can control what is outputted. we can huddle so i can explain better

mad-raccoon: ignore the above - this would work if we were using different types

src/components/charts/shared/ChartCard/ChartCard.tsx
```diff
@@ -68,6 +67,10 @@ export const ChartCard = React.forwardRef<HTMLDivElement, ChartCardProps>(
     const theme: Theme = useTheme() as Theme;
     i18nSetup(theme);
 
+    const resolvedTitle = title ? resolveI18nString(title) : title;
+    const resolvedDescription = description ? resolveI18nString(description) : description;
+    const resolvedTooltip = tooltip ? resolveI18nString(tooltip) : tooltip;
+
```
mad-raccoon: could we follow the same pattern that we already have?

```
const {tooltip, title, description} = resolved(props)
const {...} = props
```

mad-raccoon (approved):

PR #220, opened 2026-06-19
[RUI-292]: Add support for or operator in filter builder

alicjakujawa:

The FilterBuilder currently combines all clauses using AND logic only — every clause must be satisfied for a record to be included. There is no way for end users to switch between AND and OR at the top level.

# Main changes

* Wired FilterBuilderProAndOrButton between every pair of adjacent filter clauses; the operator is persisted in FilterBuilderState and emitted in the output FilterBuilderClause
* Added operator field to FilterBuilderState; defaults to AND
* Added unit tests covering the disabled state, tooltip display, and the mixed-type auto-reset behaviour

# Test evidence

https://github.com/user-attachments/assets/358e1e0e-6456-4159-b7ca-13b66dcd484d

## Summary by CodeRabbit

* **New Features**
  * Added "or" operator support for filter combinations in the filter builder.
  * Automatically switches to "and" when mixing metrics and attributes.
  * Displays a tooltip explaining why "or" is unavailable for mixed types.

* **Tests**
  * Enhanced test coverage for operator toggle behavior and mixed-type filtering scenarios.

### DIFF
--- a/src/components/editors/FilterBuilderPro/components/FilterBuilderProAndOrButton.tsx
+++ b/src/components/editors/FilterBuilderPro/components/FilterBuilderProAndOrButton.tsx
@@ -42,7 +42,7 @@ export const FilterBuilderProAndOrButton: FC<FilterBuilderProAndOrButtonProps> =
 
   return (
     <Tooltip side="top" trigger={<span>{button}</span>}>
-      {i18n.t('editors.filterBuilder.orDisabledMixedTypes')}
+      {i18n.t('editors.filterBuilder.disableOrOperatorToolTip')}
     </Tooltip>
   );
 };
--- a/src/components/editors/FilterBuilderPro/components/FilterBuilderProAndOrButton.tsx
+++ b/src/components/editors/FilterBuilderPro/components/FilterBuilderProAndOrButton.tsx
@@ -16,7 +16,6 @@ export const FilterBuilderProAndOrButton: FC<FilterBuilderProAndOrButtonProps> =
   disabled = false,
 }) => {
   const handleChange = () => {
-    if (disabled) return;
     onChange(
       operator === filterBuilderAndOrOperator.AND
         ? filterBuilderAndOrOperator.OR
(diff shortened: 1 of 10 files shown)


src/components/editors/FilterBuilderPro/components/FilterBuilderProAndOrButton.tsx
```diff
@@ -24,12 +28,27 @@ export const FilterBuilderProAndOrButton: FC<FilterBuilderProAndOrButtonProps> =
...
   const inactiveLabel = operator === filterBuilderAndOrOperator.AND ? orLabel : andLabel;
 
-  return (
-    <button className={styles.andOrButton} onClick={handleChange}>
+  const button = (
+    <button
+      className={styles.andOrButton}
+      onClick={handleChange}
+      aria-disabled={disabled}
+      data-disabled={disabled || undefined}
```
mad-raccoon: Suggest switching to native `disabled` + a wrapper `<span>` for the tooltip anchor instead of `aria-disabled` + `data-disabled`.

The current pattern keeps the button interactive, which means `handleChange` needs a manual guard (`if (disabled) return`) to prevent clicks from firing — easy to miss in future refactors. Native `disabled` lets the browser enforce that for free.

The reason to avoid native `disabled` alone is valid (pointer events are blocked, so tooltips won't trigger on hover), but that's solved cleanly by wrapping with a `<span>` and attaching the tooltip there instead.

```tsx
<span /* tooltip trigger */>
  <button
    className={styles.andOrButton}
    onClick={handleChange}
    disabled={disabled}
  >
    AND
  </button>
</span>
```

This gives you correct semantics, no click guard needed, and the tooltip still works.

src/components/editors/FilterBuilderPro/components/FilterBuilderProAndOrButton.tsx
```diff
@@ -24,12 +28,27 @@ export const FilterBuilderProAndOrButton: FC<FilterBuilderProAndOrButtonProps> =
...
     </button>
   );
+
+  if (!disabled) {
+    return button;
+  }
+
+  return (
+    <Tooltip side="top" trigger={button}>
+      {i18n.t('editors.filterBuilder.orDisabledMixedTypes')}
```
mad-raccoon: do you think there is a better naming for this key ?

mad-raccoon (changes requested):

mad-raccoon (approved):

PR #231, opened 2026-07-03
[RUI-239]: Integrate updated PivotTable

alicjakujawa:

The pivot table's subtotal configuration was limited to boolean on/off toggles per measure. The new remarkable-ui PivotTable API supports multiple aggregation types (sum, min, max, average) per axis, so the sub-inputs need to reflect that — letting users choose which aggregations to show rather than just whether to show a total.

# Main changes

Added PivotAggregation.type.emb.ts — new embeddable type with sum, min, max, average options, following the same pattern as ExportOption
Replaced showColumnTotal / showRowTotal boolean sub-inputs on measures with columnAggregation / rowAggregation multiselect sub-inputs
Replaced getPivotColumnTotalsFor / getPivotRowTotalsFor utils with getPivotColumnAggregationsFor / getPivotRowAggregationsFor, which map selected aggregation types to the individual remarkable-ui props (columnSumFor, columnMinFor, etc.)
Added onCellClick prop to HeatMap in remarkable-ui (type + implementation), which was wired up in HeatMapPro but missing from the UI library

# Test evidence

https://github.com/user-attachments/assets/adc2f242-3d6a-4687-9de2-0441622402f1

## Summary by CodeRabbit

* **New Features**
  * Pivot tables now support row and column calculations with options for sum, min, max, and average.
  * Aggregate labels in pivot tables are now translatable in supported languages.

* **Bug Fixes**
  * Updated pivot table wording from “Show row/column aggregation” to “Show row/column calculation” for clearer labeling.

### DIFF
--- a/.changeset/olive-pandas-cheer.md
+++ b/.changeset/olive-pandas-cheer.md
@@ -1,0 +1,5 @@
+---
+'@embeddable.com/remarkable-pro': patch
+---
+
+Pivot table: renamed the "Show row/column aggregation" inputs to "Show row/column calculation", and the Sum/Min/Max/Average total labels shown in the rendered table are now translatable via the theme i18n (`charts.pivotTable.sum|min|max|average`).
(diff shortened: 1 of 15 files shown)


mad-raccoon (approved):
🚀

PR #238, opened 2026-07-10
[RUI-300] Fix 'other' calculation for charts

alicjakujawa:

The chart "Other" bucket was calculated on the front-end from only the returned rows. When a query hit its row limit (component maxResults or a Cube cap), those rows were a subset, so "Other" was computed from an incomplete tail and came out understated/wrong. It should reflect the full result set.

# Main changes

Added a parallel no-dimension "grand total" query for additive (sum/count) measures.
groupTailAsOther now computes Other = grandTotal − sum(head) when a total is present (correct whether or not the set is truncated).
Ordered the results query by the first measure so the top-N head is deterministic.
Hold the chart's loading state until the total resolves, so the totals-less value never flashes.
Wired into BarChartDefaultPro, PieChartPro, DonutChartPro, DonutLabelChartPro, and BarLineChartPro; avg/min/max/median stay on the existing front-end path.

# Test evidence

https://github.com/user-attachments/assets/c05908e7-7646-4a39-9bbf-36a6f1d60234

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected “Other” calculations in bar, line, pie, and donut charts.
  * “Other” values now use full-dataset totals where available, improving accuracy when charts limit displayed items.
  * Charts now sort displayed items consistently and show a loading state while totals are being calculated.

* **Tests**
  * Added coverage for additive and non-additive measures, total calculations, sorting, loading states, and chart data handling.

### DIFF
--- a/src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
+++ b/src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
@@ -10,7 +10,7 @@ import { useFillGaps } from '../../charts.fillGaps.hooks';
 import { ChartGranularitySelectField } from '../../shared/ChartGranularitySelectField/ChartGranularitySelectField';
 import { BarChartBaseProps } from '../bars.types';
 import { createSimpleClickHandler } from '../../charts.utils';
-import { getMeasureTotals, getResultsForCard } from '../../charts.other.loadData.utils';
+import { getChartCardData } from '../../charts.other.loadData.utils';
 
 export type BarChartDefaultHorizontalProProps = BarChartBaseProps & {
   reverseYAxis?: boolean;
@@ -48,18 +48,15 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps) 
     dimension,
   });
 
-  const cardData = getResultsForCard(results, resultsOtherTotal);
+  const cardData = getChartCardData({
+    results,
+    resultsOtherTotal,
+    dimension,
+    measures,
+    maxItems: yAxisMaxItems,
+  });
 
-  const data = getBarChartProData(
-    {
-      data: results.data,
-      dimension,
-      measures,
-      maxItems: yAxisMaxItems,
-      measureTotals: getMeasureTotals(resultsOtherTotal, measures),
-    },
-    theme,
-  );
+  const data = getBarChartProData({ data: cardData.data, dimension, measures }, theme);
 
   const options = mergician(
     getBarChartProOptions({ measures, horizontal: true, data, dimension }, theme), // Format X axis based on first measure
--- a/src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
+++ b/src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
@@ -10,6 +10,7 @@ import { useFillGaps } from '../../charts.fillGaps.hooks';
 import { ChartGranularitySelectField } from '../../shared/ChartGranularitySelectField/ChartGranularitySelectField';
 import { BarChartBaseProps } from '../bars.types';
 import { createSimpleClickHandler } from '../../charts.utils';
+import { getMeasureTotals, getResultsForCard } from '../../charts.other.loadData.utils';
 
 export type BarChartDefaultHorizontalProProps = BarChartBaseProps & {
   reverseYAxis?: boolean;
@@ -36,6 +37,7 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps) 
     granularity,
     setGranularity,
     onBarClicked,
+    resultsOtherTotal,
   } = props;
 
   const resolvedI18nProps = resolveI18nProps(props);
@@ -46,8 +48,16 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps) 
     dimension,
   });
 
+  const cardData = getResultsForCard(results, resultsOtherTotal);
+
   const data = getBarChartProData(
-    { data: results.data, dimension, measures, maxItems: yAxisMaxItems },
+    {
+      data: results.data,
+      dimension,
+      measures,
+      maxItems: yAxisMaxItems,
+      measureTotals: getMeasureTotals(resultsOtherTotal, measures),
+    },
     theme,
   );
 
@@ -67,7 +77,7 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps) 
 
   return (
     <ChartCard
-      data={results}
+      data={cardData}
       dimensionsAndMeasures={[dimension, ...measures]}
       errorMessage={results.error}
       {...asChartCardHeaderProps(props)}
--- a/src/components/charts/bars/BarChartDefaultPro/index.tsx
+++ b/src/components/charts/bars/BarChartDefaultPro/index.tsx
@@ -10,12 +10,16 @@ import { useFillGaps } from '../../charts.fillGaps.hooks';
 import { ChartGranularitySelectField } from '../../shared/ChartGranularitySelectField/ChartGranularitySelectField';
 import { BarChartBaseProps } from '../bars.types';
 import { createSimpleClickHandler } from '../../charts.utils';
+import { DataResponse } from '@embeddable.com/core';
+import { getMeasureTotals, isResultTruncated } from '../../charts.other.loadData.utils';
 
 export type BarChartDefaultProProps = BarChartBaseProps & {
   reverseXAxis?: boolean;
   xAxisMaxItems?: number;
+  maxResults?: number;
   yAxisRangeMin?: number;
   yAxisRangeMax?: number;
+  resultsOtherTotal?: DataResponse;
 };
 
 const BarChartDefaultPro = (props: BarChartDefaultProProps) => {
@@ -26,6 +30,7 @@ const BarChartDefaultPro = (props: BarChartDefaultProProps) => {
     measures,
     yAxisRangeMin,
     xAxisMaxItems,
+    maxResults,
     yAxisRangeMax,
     showLegend,
     showTooltips,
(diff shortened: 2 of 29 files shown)


src/components/charts/charts.other.loadData.utils.ts
```diff
@@ -0,0 +1,90 @@
...
+  OrderBy,
+  QueryFilter,
+  loadData,
+} from '@embeddable.com/core';
+import { SortDirectionTypeOptions } from '../types/SortDirection.type.emb';
+
+const ADDITIVE_AGG_TYPES = new Set(['sum', 'count']);
+
+export const getMeasureAggType = (measure: Measure): string | undefined =>
+  (measure.meta as Record<string, unknown> | undefined)?.aggType as string | undefined;
```
mad-raccoon: lets make it less redundant 

```
export const getMeasureAggType = (measure: Measure): string | undefined =>
  measure.meta?.aggType as string | undefined;
```

src/components/charts/charts.other.loadData.utils.ts
```diff
@@ -0,0 +1,90 @@
...
+};
+
+type OtherTotalLoadDataArgs = {
+  dataset: Dataset;
+  measures: Measure[];
+  timezone?: string;
+  filters?: QueryFilter[];
+};
+
+export const otherTotalLoadDataArgs = ({
```
mad-raccoon: this names are only used for our clients to reuse our functions - in this case our users will not reuse it - so we can avoid this pattern:



```
    loadDataArgs,
    loadData,
```

just create a function that returns what it needs, and choose a good name, so it is clear that that function load something

src/components/charts/charts.other.loadData.utils.ts
mad-raccoon: also, if this is only used once, you can move it inside the isAdditiveMeasure - it is only 1 line of code

src/components/charts/charts.other.loadData.utils.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export const isAdditiveMeasure = (measure: Measure): boolean => {
+  const aggType = getMeasureAggType(measure);
+  return aggType == null || ADDITIVE_AGG_TYPES.has(aggType);
+};
+
+export const getAdditiveMeasures = (measures: Measure[]): Measure[] =>
+  measures.filter(isAdditiveMeasure);
+
+export const getTopItemsOrderBy = (measures: Measure[]): OrderBy[] => {
```
mad-raccoon: what is topItems ?


can we rename this to getFirstMeasureOrderBy

src/components/charts/charts.other.loadData.utils.ts
```diff
@@ -0,0 +1,90 @@
...
+type LoadOtherTotalArgs = OtherTotalLoadDataArgs & {
+  maxItems?: number;
+};
+
+export const loadOtherTotal = ({
+  dataset,
+  measures,
+  maxItems,
+  timezone,
+  filters,
```
mad-raccoon: this filters is never used. do we need this ?

src/components/charts/charts.other.loadData.utils.ts
```diff
@@ -0,0 +1,90 @@
...
+  loadData,
+} from '@embeddable.com/core';
+import { SortDirectionTypeOptions } from '../types/SortDirection.type.emb';
+
+const ADDITIVE_AGG_TYPES = new Set(['sum', 'count']);
+
+export const getMeasureAggType = (measure: Measure): string | undefined =>
+  (measure.meta as Record<string, unknown> | undefined)?.aggType as string | undefined;
+
+export const isAdditiveMeasure = (measure: Measure): boolean => {
```
mad-raccoon: why are we creating one function each time we need something ?

this functions are so simple that they dont need to be defined

this creates more  code, and more required tests

src/components/charts/charts.other.loadData.utils.ts
```diff
@@ -0,0 +1,90 @@
...
+  measures: Measure[];
+  timezone?: string;
+  filters?: QueryFilter[];
+};
+
+export const otherTotalLoadDataArgs = ({
+  dataset,
+  measures,
+  timezone,
+  filters,
```
mad-raccoon: filters is never passed

mad-raccoon (changes requested):

src/components/charts/charts.other.loadData.utils.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+type LoadOtherTotalArgs = {
+  dataset: Dataset;
+  measures: Measure[];
+  maxItems?: number;
+  timezone?: string;
+};
+
+export const loadOtherTotal = ({
```
mad-raccoon: can we rename this to loadDataOtherTotal ?

because loadData is the SDK function - is more clear it relates to that

src/components/charts/bars/BarChartDefaultPro/index.tsx
```diff
@@ -10,12 +10,16 @@ import { useFillGaps } from '../../charts.fillGaps.hooks';
...
+import { DataResponse } from '@embeddable.com/core';
+import { getMeasureTotals, getResultsForCard } from '../../charts.other.loadData.utils';
 
 export type BarChartDefaultProProps = BarChartBaseProps & {
   reverseXAxis?: boolean;
   xAxisMaxItems?: number;
+  maxResults?: number;
   yAxisRangeMin?: number;
   yAxisRangeMax?: number;
+  resultsOtherTotal?: DataResponse;
```
mad-raccoon: if i am not mistaken, this can be moved inside the BarChartBaseProps - as it needs to be available for all the barcharts, not ony the BarChartDefaultProProps

mad-raccoon: for example, the BarChartDefaultHorizontalPro also needs this

src/components/charts/bars/BarChartDefaultPro/index.tsx
```diff
@@ -46,8 +51,14 @@ const BarChartDefaultPro = (props: BarChartDefaultProProps) => {
     dimension,
   });
 
+  const otherOptions = {
+    measureTotals: getMeasureTotals(resultsOtherTotal, measures),
+  };
```
mad-raccoon: i think we can simplify this avoiding a nested object 

also, otherOptions is very vague

my suggestion: drop the nesting, and just pass measureTotals

```
 { data: results.data, dimension, measures, maxItems: xAxisMaxItems,  measureTotals: getMeasureTotals(resultsOtherTotal, measures) },
```

src/components/charts/charts.utils.ts
```diff
@@ -10,11 +10,16 @@ export const getDimensionWithoutTruncation = (dimension: Dimension): Dimension =
...
+export type GroupTailAsOtherOptions = {
+  measureTotals?: Record<string, number>;
+};
+
 export const groupTailAsOther = (
   data: DataResponse['data'] = [],
   dimension: Dimension,
   measures: Measure[],
   maxItems?: number,
+  options?: GroupTailAsOtherOptions,
```
mad-raccoon: lets create a type, but without nesting

`type MeasureTotals =  Record<string, number>`

mad-raccoon:
```suggestion
  measureTotals?: MeasureTotals,
```

src/components/charts/charts.other.loadData.utils.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (!maxItems) return undefined;
+  const additiveMeasures = getAdditiveMeasures(measures);
+  if (!additiveMeasures.length) return undefined;
+  return loadData({ from: dataset, select: additiveMeasures, timezone });
+};
+
+export const getMeasureTotals = (
+  otherTotalResults: DataResponse | undefined,
+  measures: Measure[],
+): Record<string, number> => {
```
mad-raccoon: lets use the type **MeasureTotals**

mad-raccoon (changes requested):

src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
```diff
@@ -46,8 +48,16 @@ const BarChartDefaultHorizontalPro = (props: BarChartDefaultHorizontalProProps)
     dimension,
   });
 
+  const cardData = getResultsForCard(results, resultsOtherTotal);
```
mad-raccoon: **question:** what is supposed to happen when the user clicks on the download-button ?

should it download all the data ? or still show the top x measures, and then the "other"


e.g. on the paginated table, we have a separated loadData specific for the download - when the user clicks download, on the paginated table, we fetch all the data on the background, so when it downloads, it has all the data.

(otherwise it would only have the results of the current visible page)

we should check with harry this

mad-raccoon (commented):
**Question:** shouldnt this improvement be also done on the grouped barcharts ?

mad-raccoon (changes requested):
other than those 2 remarks looks good!

i would check the download part -  also because the function name **getResultsForCard** is too generic

src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
alicjakujawa: @hjkmarshall

hjkmarshall: @alicjakujawa let's make it download the data it's actually showing - i.e. append the 'other' load data result to the first load data call results. Sound ok?

alicjakujawa:
> **Question:** shouldnt this improvement be also done on the grouped barcharts ?

they dont have other aggregation at all

src/components/charts/bars/BarChartDefaultHorizontalPro/index.tsx
mad-raccoon: @alicjakujawa if this is already how it works, lets just find a better naming for the function 😎

mad-raccoon (approved):

PR #237, opened 2026-07-09
[RUI-296]: Introduce separate component for filter grouping

alicjakujawa:
# Why is this pull-request needed?

Users currently can't group filter conditions with mixed AND/OR logic (e.g. Age > 10 AND (Date is X OR Country is Y) AND Count > 5) — filters are a flat list. This adds nested filter groups, matching the Intercom-style UX from design. To avoid any risk to existing customers (who may have themed or rely on today's FilterBuilderPro), the new grouping behavior ships as a separate, additive component rather than modifying the existing one in place.

# Main changes

New component FilterBuilderWithGrouping — flat filters plus single-level nested AND/OR groups, Cube-safe (a group can't mix dimensions and measures; the top-level OR is disabled when types are mixed), with new two-tone segment styling, group container, overflow chevron, and per-filter delete/create-group actions per the updated design.
FilterBuilderPro unchanged — reverted to its original behavior, tokens, and output; zero impact to existing customers or themes.
Shared logic extracted to editors/utils/filterBuilder.utils.ts (operator maps, clause serialization, value helpers) — reused by both components to avoid duplication.
Value/operator editor shared via parameterization (optional styles + showInlineClear) — one implementation renders old design for FilterBuilderPro and new design for FilterBuilderWithGrouping.
Additive theme tokens (--em-filter-*) and i18n keys — legacy --em-filterbuilder-* tokens untouched.

# Test evidence

https://github.com/user-attachments/assets/3733cf12-b958-425b-bf07-8add3e19b668

## Summary by CodeRabbit

* **New Features**
  * Added a grouped filter builder experience with support for nested filter groups, AND/OR toggles, adding/removing filters, and creating groups.
  * Added updated styling and translation text for the new filter builder UI.
  * Added support for customizable filter field styling and clearer inline controls.

* **Bug Fixes**
  * Improved filter-state handling so grouped filters stay in sync and invalid operator combinations are automatically corrected.

* **Tests**
  * Added coverage for grouped filter behavior, operator toggling, option filtering, scrolling, and state conversion.

PR #249, opened 2026-07-20
Adopt defaultFilters changes after mount in filter builders

hjkmarshall:
## Context
A client reported that FilterBuilderPro treats its defaultFilters input as a one-time seed, not a trackable value. Once the component holds any internal filter state, later host-driven changes to the bound variable are dropped — hosts can seed the value but can't programmatically update or reset it afterward. Their only workaround was remounting the entire <em-beddable> element (bumping its key), which drops all other dashboard state.

Root cause — two layers:

Seed-once guard: the seeding effect bailed out via if (prev?.filters?.length) return prev. This guard existed for a real reason — defaultFilters and onChange are bound to the same platform variable, so every user edit echoes back in as a "new" defaultFilters, and a naive "always adopt" would clobber edits.
Stale value inputs (the deeper one): even once the state is updated, the value inputs (FilterBuilderItemNumberValueField / FilterBuilderTextValueField) seed their own local useState from filter.value only on mount and never re-sync. So adopting a new value into embeddableState updated the state but the rendered input kept showing the stale value. This is why an adopt-into-state approach alone looked broken in a real embed while unit tests (which mock the value inputs) passed.
The newly-added FilterBuilderWithGroupingPro ([#237](https://github.com/embeddable-hq/remarkable-pro-boilerplate/issues/237)) carried the identical behaviour, so both are addressed here.

## What changed
New opt-in input syncDefaultFilters ("Track default filters", default false). Off → existing seed-once behaviour, unchanged (safe as a patch). On → the filter tracks host-driven changes to defaultFilters after mount.
New shared hook useAdoptDefaultFilters in filters.hooks.ts — adopts defaultFilters on mount and, when syncing, whenever the host pushes a genuinely new value (switching between / reverting saved filter sets, or resetting).
Echo-immune: ignores any incoming value equal to what the component last emitted (via the existing prevFilterValueRef) or last adopted, so mid-session edits are never overwritten.
Reset semantics: null / undefined / {} / sentinel objects (e.g. Value.noFilter()) are ignored so a missing binding never wipes state; a well-formed empty clause { operator, clauses: [] } is honoured as an explicit reset to a single blank filter.
Value inputs now reflect adopted changes: an adoptRevision counter is mixed into the filter-row key while syncing, so an adopted change remounts the row and the value inputs re-read the new filter.value — fixing layer 2 without reworking the debounced-input UX.
FilterBuilderPro also restores the saved top-level AND/OR operator on adopt, matching the grouping variant.
Both components share the hook and the same flag.

## Testing
Added a new FilterBuilderItemNumberValueField.test.tsx that reproduces the stale-input bug (a same-instance filter.value change is not reflected) and proves the fix mechanism (a key change / remount re-seeds it) — the coverage the mocked component tests structurally couldn't provide.
Extended both components' index.test.tsx to cover flag-off (seed-once) and flag-on (adopt / reset / echo-immunity / null-ignore) paths.
258/258 tests pass across src/components/editors/filters; typecheck and ESLint clean.
Verified end-to-end in a live embed: flag off → seed-once (pushes ignored); flag on → a host push updates the visible filter (67 → 100), an empty clause resets it, and null is ignored (current value preserved).

## Notes
The input is a backward-compat / opt-in switch, not a new host-control axis — the host already controls updates by whether it pushes a new variable value. It exists so existing dashboards keep seed-once behaviour and this can ship as a patch. If we later decide variable-tracking should be the default, the follow-up is to make the value inputs genuinely controlled (sync filter.value without the focus-losing remount), then flip the default and retire the flag.
While syncing, an adopted change remounts the filter row, which drops input focus — acceptable for a programmatic set/reset, but the reason above for keeping it opt-in.
If the client's host represents "reset to empty" with Value.noFilter() or a bare {} rather than { operator, clauses: [] }, that won't trigger the reset — worth confirming; extending the guard is a one-liner.

## Evidance

https://github.com/user-attachments/assets/744f8bcc-e7e1-4d08-8dd3-89383a87ee7f

### DIFF
--- a/src/components/editors/filters/FilterBuilderPro/index.tsx
+++ b/src/components/editors/filters/FilterBuilderPro/index.tsx
@@ -38,6 +38,7 @@ export type FilterBuilderProProps = {
   dimensionsAndMeasures?: DimensionOrMeasure[];
   onChange?: (value: unknown) => void;
   defaultFilters?: FilterBuilderClause;
+  syncDefaultFilters?: boolean;
 } & EditorCardHeaderProps;
 
 const FilterBuilderPro = (props: FilterBuilderProProps) => {
@@ -51,23 +52,34 @@ const FilterBuilderPro = (props: FilterBuilderProProps) => {
     embeddableState,
     onChange,
     defaultFilters,
+    syncDefaultFilters = false,
   } = props;
 
   const [searchNew, setSearchNew] = useState('');
   const prevFilterValueRef = useRef<unknown>(undefined);
+  // Bumped each time an adopted value is applied. Mixed into the filter row keys
+  // so the value inputs (which seed their own local state on mount) remount and
+  // re-read the adopted value. Only used when syncDefaultFilters is on.
+  const [adoptRevision, setAdoptRevision] = useState(0);
 
-  // Adopt defaultFilters on mount and whenever the host pushes a genuinely new
-  // value. An empty clause list resets to a single blank filter (the render
-  // fallback below turns `filters: []` into one empty row).
   const adoptDefaultFilters = useCallback(
     (clause: FilterBuilderClauseGroup) => {
-      setEmbeddableState?.((prev) => ({
-        ...prev,
-        filters: clauseToFilters(clause, dimensionsAndMeasures),
-        operator: clause.operator,
-      }));
+      setEmbeddableState?.((prev) => {
+        // Seed-once (default, backward compatible): only apply while empty.
+        if (!syncDefaultFilters && prev?.filters?.length) {
+          return prev;
+        }
+        return {
+          ...prev,
+          filters: clauseToFilters(clause, dimensionsAndMeasures),
+          operator: clause.operator,
+        };
+      });
+      if (syncDefaultFilters) {
+        setAdoptRevision((revision) => revision + 1);
+      }
     },
-    [dimensionsAndMeasures, setEmbeddableState],
+    [dimensionsAndMeasures, setEmbeddableState, syncDefaultFilters],
   );
 
   useAdoptDefaultFilters({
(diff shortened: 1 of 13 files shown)


src/components/editors/filters/FilterBuilderWithGroupingPro/definition.ts
```diff
@@ -38,6 +38,15 @@ const meta = {
...
     },
+    {
+      ...inputs.boolean,
+      name: 'syncDefaultFilters',
+      label: 'Track default filters',
+      defaultValue: false,
+      category: 'Component Settings',
+      description:
+        'When on, the filter tracks changes to its bound Default filters after load — so the host can update or reset it at runtime. When off (default), Default filters only seeds the initial value.',
+    },
```
mad-raccoon: code duplication.

if reused, lets create a var for it

src/components/editors/filters/FilterBuilderPro/index.tsx
```diff
@@ -50,31 +55,45 @@ const FilterBuilderPro = (props: FilterBuilderProProps) => {
...
     componentName,
     trackingId,
   } = props;
 
   const [searchNew, setSearchNew] = useState('');
   const prevFilterValueRef = useRef<unknown>(undefined);
+  // Bumped each time an adopted value is applied. Mixed into the filter row keys
+  // so the value inputs (which seed their own local state on mount) remount and
+  // re-read the adopted value. Only used when syncDefaultFilters is on.
+  const [adoptRevision, setAdoptRevision] = useState(0);
```
mad-raccoon: The stale value comes from the two value fields copying `filter.value` into local state on mount and never looking again. `adoptRevision` works around that by remounting rows via the `key`.

Fixing the fields directly looks like less code - same echo trick you already wrote, just on a scalar:

```ts
useEffect(() => {
  const incoming = (filter?.value as number) ?? null;
  if (incoming === lastEmittedRef.current) return;  // our own echo
  setValue(incoming);
}, [filter.value]);
```

That drops the counter, the `key` ternary and probably the flag, and fixes it for every consumer of those fields.

I haven't tried it though - controlled + debounced + updated from outside is fiddly. If this does the trick it is preferred than this workaround.

mad-raccoon (changes requested):
on another note: `syncDefaultFilters` is checked in three places across two files, never in the hook. So the hook decides "did it change?" and the component decides "do we care?", twice. No single place to read the answer, which is what made this hard to follow. I'd have the hook own the flag and the callback just map a clause into state.

i think this is clear - claude will help you on how to approach this.

mad-raccoon (approved):

PR #253, opened 2026-07-29
[RUI-304]: Add possbility to connect lines when empty dates

alicjakujawa:

Line/area/combo charts use fillGaps to insert placeholder rows for dates with no data, so charts render a bucket for every date in the range rather than only for dates with rows. Today, when "Ignore empty dates" is off, those placeholder rows carry a date but no measure value — which Chart.js renders as a break in the line (disconnected dots), with no way for the customer to opt into a connected line instead.

This was reported via [RUI-304], reproduced with a Line Chart Comparison where "Ignore empty dates" was off: the load-data response had real gaps between dates (e.g. no rows for Jul 2–6), and the chart showed disconnected segments instead of a continuous line.

This PR adds a "Connect gaps" input so customers can choose to connect the line across empty dates by treating the missing value as 0, instead of leaving a visible break.

# Main changes

Added a new Connect gaps boolean input (default false, preserving current behavior) on the x-axis date dimension of LineChartDefaultPro, LineChartGroupedPro, LineChartComparisonDefaultPro, and BarLineChartPro (and their variants that inherit these definitions: AreaChartPro, LineChartWithKpiTabsPro, LineChartComparisonWithKpiTabsPro). It's gated to time dimensions (supportedTypes: ['time']), so it only appears alongside the existing Ignore empty dates input and never on non-line charts.
When enabled, each chart's data-transform util fills missing measure values with 0 instead of null for that bucket, so Chart.js draws a continuous line across the gap instead of breaking it.
The setting lives on the date dimension (not the measure) because it governs the shared x-axis behavior for the whole chart, matching how ignoreEmptyDate already works.
Added src/components/charts/charts.fillGaps.hooks.test.ts, a reproduction test using a dataset modeled on the reported gappy series, covering: the root-cause placeholder behavior in useFillGaps, the ignoreEmptyDate interaction, and the connected-vs-disconnected line output end-to-end.
Updated existing BarLineChartPro and LineChartDefaultPro util tests to reflect the input moving from measure to dimension.

# Test evidence

https://github.com/user-attachments/assets/2a203bc7-a073-4d52-b1ef-3df6dc39ad47

[RUI-304]: https://trevorio.atlassian.net/browse/RUI-304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added a **Connect gaps** option for line charts.
  * When enabled, missing dates are filled with zeroes so chart lines remain continuous.
  * Applied the option consistently across standard, grouped, comparison, and bar-line charts.
* **Bug Fixes**
  * Corrected gap handling when dates are missing, preserving visible breaks when the option is disabled.
* **Tests**
  * Added coverage for missing-date and line-connection behavior.


mad-raccoon (changes requested):
this one was already implemented after all
the connectgaps was inside of the measures

we should not change this , otherwise clients that use this will be affected

please test with the measure.connectGaps - if it works there is no work needed

mad-raccoon:
left evidence here: https://trevorio.atlassian.net/browse/RUI-304

PR #264, opened 2026-08-21
DO NOT MERGE Allows fillGaps to be Time Zone Aware

cwbuecheler:

One of our customers, Orb, has determined that there is a problem with the `fillGaps` function when used in conjunction with time zones and date-bound charts. It's returning the correct data, but in UTC rather than in the appropriate timezone. This causes the chart to render incorrectly, often showing half a day of data and then 0s beyond, as those UTC date/time combinations haven't actually happened in, say, PST yet. This only happens with date-bound charts.

  1. I query data for now (let's say 11 AM) to one day earlier than now in PST3. 
  2. The query correctly returns that data - I get 11 AM yesterday to 11 AM today, PST. No problems there.
  3. `fillGaps` runs, and converts the date bounds to UTC
  4. The last seven hours of the chart are empty because "now" has been converted 11 AM UTC instead of 11 AM PST, but "now" is only 4 AM PST. So the chart fills until 4 AM and then stops.

Basically the issue is that the data's there but it's getting skewed by `UTC - timezone` hours because fillGaps wasn't time zone aware.

Pictured: One chart that's not date bound and is working correctly, and below it one that's displaying the issue:

image

# Main changes

Leverages DayJS's time zone functionality to make fillGaps time zone aware

# Test evidence

UTC:
image

PST:
image

Note that the 98k value correctly moves from 1 AM August 16 in UTC to 5 PM August 15 in PST.

You can ignore the zeroes at the end of the chart. That's related to the data, not a display issue as mentioned above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chart date handling across time zones.
  * Date ranges now correctly interpret calendar dates in the client’s local time zone.
  * Added safer handling for missing or fallback date values when filling chart gaps.


mad-raccoon:
@cwbuecheler looks good. 

my only concern is if this will affect something when the granularity is not hour (is week or day)

cwbuecheler:
> @cwbuecheler looks good.
> 
> my only concern is if this will affect something when the granularity is not hour (is week or day)

Ugh, yep, it doesn't work in PST at all if you set the granularity to day. UTC works fine. I'll keep messing with it.

mad-raccoon:
> > @cwbuecheler looks good.
> > my only concern is if this will affect something when the granularity is not hour (is week or day)
> 
> Ugh, yep, it doesn't work in PST at all if you set the granularity to day. UTC works fine. I'll keep messing with it.

i think the solution we need is something identical to this:

```
const getLocalDateString = (date: Date, tz?: string): string =>
  tz ? dayjs.tz(date, tz).format('YYYY-MM-DD') : dayjs.utc(date).format('YYYY-MM-DD');
```

this was implemented on the defaults.DateRange.constants.ts file

cwbuecheler:
@mad-raccoon - Bug is fixed and verified, also fixed the bug that coderabbit found. I also added test coverage.

cwbuecheler:
@mad-raccoon - FYI I've tested this with Olivier and it doesn't appear to be working correctly for his use case. I'm going to take another stab at it in a separate branch, using your time zone work on the date picker as reference.

cwbuecheler:
Closing this and deleting the branch. Customer confirms the other one works better.

PR #269, opened 2026-08-26
CS-282: Create FunnelChart Component

miguelopezv:

Add a FunnelChart component to the library.

# Main changes

- New FunnelChart component built on chartjs-chart-funnel, with support for legend, tooltips, count, and percentage labels
- Added tests and a Storybook story
- Exported FunnelChart from the package entry point
- Added changeset (minor) for the new component

## Summary by CodeRabbit

* **New Features**
  * Added the FunnelChartPro component for visualizing stage-based conversion funnels.
  * Added configurable sorting, colors, gradients, legends, tooltips, percentages, titles, descriptions, and menu options.
  * Added predefined funnel color schemes and theme-specific chart configuration.
  * Exported funnel chart types, utilities, and component configuration for integration.

* **Tests**
  * Added coverage for funnel data handling, rendering, theming, formatting, colors, and edge cases.


src/components/types/FunnelColorScheme.type.emb.ts
```diff
@@ -0,0 +1,30 @@
...
+const FunnelColorSchemeType = defineType('funnelColorScheme', {
+  label: 'Color scheme',
+  optionLabel: (value: string) => colorSchemeLabelMap[value] ?? value,
+});
+
+defineOption(FunnelColorSchemeType, FunnelColorSchemeOptions.amber);
+defineOption(FunnelColorSchemeType, FunnelColorSchemeOptions.orange);
+defineOption(FunnelColorSchemeType, FunnelColorSchemeOptions.blue);
+defineOption(FunnelColorSchemeType, FunnelColorSchemeOptions.greenRed);
+defineOption(FunnelColorSchemeType, FunnelColorSchemeOptions.red);
```
mad-raccoon: are this options a requirement ?

how our charts work is by accessing the getChartColors

```
export const getChartColors = (): string[] => [
  getStyle('--em-sem-chart-color--1'),
  getStyle('--em-sem-chart-color--2'),
  getStyle('--em-sem-chart-color--3'),
  getStyle('--em-sem-chart-color--4'),
  getStyle('--em-sem-chart-color--5'),
  getStyle('--em-sem-chart-color--6'),
  getStyle('--em-sem-chart-color--7'),
  getStyle('--em-sem-chart-color--8'),
  getStyle('--em-sem-chart-color--9'),
  getStyle('--em-sem-chart-color--10'),
];
```

this are the colors that can be overwritten via the theming - this is doe by the user if they want to

src/components/charts/funnel/FunnelChartPro/definition.ts
```diff
@@ -0,0 +1,137 @@
...
+      description:
+        'Optional numeric dimension that defines stage order (ascending). When set, overrides the default descending-by-count order, e.g. severity_order INTEGER.',
+    },
+    {
+      name: 'colorScheme',
+      type: FunnelColorSchemeType,
+      label: 'Color scheme',
+      defaultValue: FunnelColorSchemeOptions.amber,
+      category: 'Component Settings',
+    },
```
mad-raccoon: unless this is a requirement, this is not required - otherwise we will be introducing a new FunnelColorSchemeType only for one single component

src/components/charts/funnel/FunnelChartPro/definition.ts
```diff
@@ -0,0 +1,137 @@
...
+      label: 'Start color (lowest stage)',
+      category: 'Component Settings',
+      description: 'Both Start and End color must be set to override the Color scheme.',
+    },
+    {
+      ...subInputs.color,
+      name: 'endColor',
+      label: 'End color (highest stage)',
+      category: 'Component Settings',
+    },
```
mad-raccoon: if the type is not a requirement, then we dont need this additional props

src/components/charts/funnel/FunnelChartPro/definition.ts
```diff
@@ -0,0 +1,137 @@
...
+    {
+      ...inputs.boolean,
+      name: 'showPercentage',
+      label: 'Show percentage',
+      description: 'Show percentage of total instead of the raw count on each stage.',
+      defaultValue: false,
+      category: 'Component Settings',
+    },
+    inputs.menuOptions,
+  ],
```
mad-raccoon, line 58: we can reuse the inputs.displayPercentages

src/components/charts/funnel/FunnelChartPro/definition.ts
```diff
@@ -0,0 +1,137 @@
...
+    ['severity', 'count'],
+    [
+      ['Near Misses', '33'],
+      ['Injury/Illness', '30'],
+      ['Recordable', '14'],
+      ['DART', '5'],
+    ],
+  ),
+  hideMenu: true,
+};
```
mad-raccoon, line 68: the preview data should be similar to all the charts - we already have some predefined data for this, check out the `preview.data.constants.ts` file

Image

src/components/charts/funnel/FunnelChartPro/definition.ts
```diff
@@ -0,0 +1,137 @@
...
+
+const preview = definePreview(Component, previewConfig);
+
+const loadDataResultsArgs = (inputs: Inputs<typeof meta>): LoadDataRequest => ({
+  from: inputs.dataset,
+  select: [
+    inputs.stageDimension as unknown as Dimension,
+    inputs.countMeasure as unknown as Measure,
```
mad-raccoon: we should not need to use unknowns in our code - check for example the `src/components/charts/pies/DonutChartPro/definition.ts` - we dont need it there for example

src/components/charts/funnel/FunnelChartPro/definition.ts
```diff
@@ -0,0 +1,137 @@
...
+
+const loadDataResults = (inputs: Inputs<typeof meta>): DataResponse =>
+  loadData(loadDataResultsArgs(inputs));
+
+const props = (inputs: Inputs<typeof meta>) => ({
+  ...inputs,
+  colorScheme: inputs.colorScheme as string | undefined,
+  startColor: inputs.startColor as string | undefined,
+  endColor: inputs.endColor as string | undefined,
+  results: loadDataResults(inputs),
```
mad-raccoon: is there a way to get rid of type assertions ?  i dont think we ever use it in our components

src/components/charts/funnel/FunnelChartPro/index.tsx
```diff
@@ -0,0 +1,67 @@
+import { useTheme } from '@embeddable.com/react';
+import { FunnelChart } from '@embeddable.com/remarkable-ui';
+import { Theme } from '../../../../theme/theme.types';
+import { getFunnelChartProData } from '../funnel.utils';
+import { DefaultFunnelChartProps } from '../funnel.types';
+import { i18nSetup } from '../../../../theme/i18n/i18n';
+import { ChartCard, asChartCardHeaderProps } from '../../shared/ChartCard/ChartCard';
+
+export type FunnelChartProProps = DefaultFunnelChartProps;
```
mad-raccoon: if there is only one chart, we dont need to specify a default type, we can create it directly here 

for the **barcharts** is different, as we have multiple variants and we dont want to duplicate code

src/components/charts/funnel/FunnelChartPro/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+      orderDimension,
+      colorScheme,
+      startColor,
+      endColor,
+    },
+    theme,
+  );
+
+  const options = theme.charts.funnelChartPro?.options ?? {};
+  const legendPosition = theme.charts.legendPosition === 'right' ? 'right' : 'bottom';
```
mad-raccoon: can we pass this inside of the prop options instead ? it will be one less dependence on this repo

src/components/charts/funnel/funnel.utils.ts
```diff
@@ -0,0 +1,97 @@
...
+import { i18n } from '../../../theme/i18n/i18n';
+import { FunnelPalette } from './funnel.types';
+
+export const FUNNEL_PALETTES: Record<string, FunnelPalette> = {
+  amber: { start: '#FBC02D', end: '#6A1A9A' },
+  orange: { start: '#FFB300', end: '#BF360C' },
+  blue: { start: '#64B5F6', end: '#0D47A1' },
+  'green-red': { start: '#66BB6A', end: '#B71C1C' },
+  red: { start: '#FFCDD2', end: '#B71C1C' },
+};
```
mad-raccoon: lets check with product/design if this colors sets are really required, as this will introduce an additional dependency (i dont think we need it)

the idea is to have a set of colors that all the chart always use - if the clients are not happy with those colors, they change them via the theme - and this will affect all the charts at once

src/components/charts/funnel/funnel.utils.ts
```diff
@@ -0,0 +1,97 @@
...
+  const palette: FunnelPalette =
+    props.startColor && props.endColor
+      ? { start: props.startColor, end: props.endColor }
+      : (FUNNEL_PALETTES[props.colorScheme ?? 'amber'] ?? FUNNEL_PALETTES.amber)!;
+
+  const backgroundColor = getColorGradient(palette.start, palette.end, stages.length);
+  const themeFormatter = getThemeFormatter(theme);
+  const dimensionWithoutTruncation = getDimensionWithoutTruncation(props.stageDimension);
+
+  return {
```
mad-raccoon, line 121: the label position should be passed here - if allowed

mad-raccoon (changes requested):
the codes looks good - there are some remarks related to product/design that we should clarify before - to avoid extra dependencies

src/components/charts/funnel/funnel.utils.ts
```diff
@@ -0,0 +1,101 @@
...
+    props.stageDimension,
+    props.countMeasure,
+    props.orderDimension,
+  );
+
+  if (!stages.length) {
+    return { labels: [], datasets: [{ data: [] }] };
+  }
+
+  const palette = resolveFunnelPalette(props.startColor, props.endColor);
```
mad-raccoon, line 110: what happens if no color is passed ?

miguelopezv: If neither startColor nor endColor is passed, resolveFunnelPalette falls through all three checks and calls getDefaultFunnelPalette() (funnel.utils.ts:16-19)

package.json
mad-raccoon: we will need to update this dependency after the RUI is released

mad-raccoon: and also run "npm run reinstall" for the package-lock

mad-raccoon:
Re: the `import type {} from 'chartjs-chart-funnel';` lines in `src/theme/theme.types.ts` and `src/components/charts/funnel/funnel.utils.ts` —

This is a type-only import, so it's fully erased at compile time (no runtime code executes). The comment above it calling it a "side-effect import" is a bit misleading.

It is still needed: `chartjs-chart-funnel` is a devDependency here, pulled in only to get its module augmentation so `'funnel'` type-checks as a valid Chart.js chart type (used in `ChartData<'funnel'>` and `ChartOptions<'funnel'>`). Actual rendering/registration of the funnel controller happens in `@embeddable.com/remarkable-ui`'s `FunnelChart`, not in this package.

Suggest tweaking the comment to avoid "side-effect", e.g.:
```ts
// Type-only: pulls in chartjs-chart-funnel's module augmentation so 'funnel' is a valid Chart.js chart type.
import type {} from 'chartjs-chart-funnel';
```

src/theme/theme.types.ts
```diff
@@ -1,4 +1,6 @@
 import { Resource } from 'i18next';
+// Side-effect import: augments Chart.js's ChartTypeRegistry with the 'funnel' chart type.
```
mad-raccoon: i dont think we need this, left a comment

src/components/charts/funnel/funnel.utils.ts
```diff
@@ -0,0 +1,101 @@
...
+import { getDimensionWithoutTruncation } from '../charts.utils';
+import { brightenColor, getColorGradient } from '../../../utils/color.utils';
+import { i18n } from '../../../theme/i18n/i18n';
+import { getChartColors } from '@embeddable.com/remarkable-ui';
+import { FunnelPalette } from './funnel.types';
+
+const PALETTE_BRIGHTEN_AMOUNT = 2.2;
+
+export const getDefaultFunnelPalette = (): FunnelPalette => {
+  const end = getChartColors()[0] ?? '#FBC02D';
```
mad-raccoon: where is the #FBC02D comming from ? we should reuse our design system token whenever we can

mad-raccoon: re. colors: will this support the colors per dimension value? 

ie. some client might want to assign a specific color to a dimension/metric value 

e.g. dimension of value **America**, will always be **red**

let me know if this is clear. you can see it working on this files:

https://github.com/embeddable-hq/remarkable-pro/blob/6ab4a76a1346ddc51b7deb276bd22164ec4971f4/src/theme/theme.types.ts#L25-L28

https://github.com/embeddable-hq/remarkable-pro/blob/6ab4a76a1346ddc51b7deb276bd22164ec4971f4/src/components/charts/bars/bars.utils.ts#L38-L45

mad-raccoon: in this case it should be our last color from the getChartColors

mad-raccoon: the color assign should be done like it is done above: the color maps take priority, falling back on the chartColors

mad-raccoon (changes requested):

src/components/charts/funnel/FunnelChartPro/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    ...theme.charts.funnelChartPro?.options,
+    plugins: {
+      ...theme.charts.funnelChartPro?.options?.plugins,
+      legend: {
+        ...theme.charts.funnelChartPro?.options?.plugins?.legend,
+        position: theme.charts.legendPosition ?? 'bottom',
+      },
+    },
+  };
+
```
mad-raccoon, line 58: lets get a function for the options, like we do on the other component. take an already existing component as guidance pls

src/components/charts/funnel/funnel.utils.ts
```diff
@@ -0,0 +1,122 @@
...
+import { ChartData } from 'chart.js';
+// Type-only: pulls in chartjs-chart-funnel's module augmentation so 'funnel' is a valid Chart.js chart type.
+import type {} from 'chartjs-chart-funnel';
+import { getThemeFormatter } from '../../../theme/formatter/formatter.utils';
+import { i18n } from '../../../theme/i18n/i18n';
+import { remarkableTheme } from '../../../theme/theme.constants';
+import { Theme } from '../../../theme/theme.types';
+import { brightenColor, getColorGradient } from '../../../utils/color.utils';
+import { getDimensionWithoutTruncation } from '../charts.utils';
+import { FunnelPalette } from './funnel.types';
```
mad-raccoon: this is only used once, we can move it here - no need for an extra file

mad-raccoon (changes requested):

src/components/charts/funnel/FunnelChartPro/index.tsx
mad-raccoon: still unresolved

lets follow the already existing patterns pls

mad-raccoon (changes requested):

mad-raccoon (approved):

PR #273, opened 2026-09-03
Improve FillGaps Time Zone Handling

cwbuecheler:
# Fix timezone handling in `useFillGaps`

## Context

Date-bound charts in a non-UTC theme timezone (e.g. `America/Los_Angeles`) show real data for part of the range, then a solid block of null buckets exactly as wide as the timezone's UTC offset (7 hours in PDT). Reported on an hourly chart bound to a live/rolling date range.

**Root cause, confirmed by reproduction against real customer data (`sample_results_2.json` not included in this PR, a 25-row, gap-free hourly Cube response for `speed.timestamp`):**

- Cube returns time-dimension values as timezone-*naive* strings — the SQL does `toTimeZone(timestamp, 'America/Los_Angeles')`, so `"2026-09-02T05:00:00.000"` means 5am *local* time, with no UTC offset in the string. `useFillGaps` (`src/components/charts/charts.fillGaps.hooks.ts`) parses these with `dayjs.utc(value)`, which is correct *only* because the string has no offset to strip — it just reads the digits as given, and those digits are already the local wall-clock reading.
- The chart's `dateBounds.to` was a **live/rolling value** (not one of the built-in day-precision presets) — its actual `Date` object holds a genuine, correctly-resolved real instant (confirmed: the query's `dateRange` filter sent to Cube, `"...T05:51:14.999"`, is the *local*-timezone string form of that same real instant, and produced a complete, gap-free result — so whatever builds that query filter already converts real instants to local time correctly).
- `useFillGaps` line 73-78 does `dayjs.utc(dateBounds.to)` on that real instant — this reads its *real UTC* digits, not its *local* digits. I reproduced this exactly with the actual sample data and dayjs (script output below): treating the bound as a real instant and converting it correctly to local digits realigns every bucket with the real records; treating it the way `useFillGaps` currently does (`dayjs.utc()` on the raw real instant) reproduces the *exact* reported failure — 18 real hours followed by 7 null hours, where 7 = the PDT UTC offset.

```
=== reproduction (dayjs, real sample_results_2.json data) ===
Treating dateBounds.to as a real instant needing local-digit conversion:
  → all 25 buckets match real records, zero nulls

Treating dateBounds.to the way useFillGaps does today (dayjs.utc() on the real instant):
  2026-09-02T12:00 .. 2026-09-03T05:00  → 18 real buckets
  2026-09-03T06:00 .. 2026-09-03T12:00  → 7 <NULL> buckets   (exactly the UTC offset)
```

- This is *not* a query-filter or backend bug — the query returned complete, correct, gap-free data. It's not fixable by touching any `definition.ts` in this repo either — none of them construct a `dateRange` filter from `dimension.inputs.dateBounds`, confirming that translation happens entirely outside this repo (platform/SDK-side), and it already behaves correctly there. The defect is entirely inside `useFillGaps`'s client-side comparison.
- Critically: **not every `dateBounds`/`externalDateBounds` value uses the same convention.** The built-in day-precision presets (`theme.defaults.dateRangesOptions`, fixed by PR #271) and the custom absolute-range picker (`getTimeRangeFromDateRange` in `dates.utils.ts`) deliberately construct a Date whose UTC-read digits already equal the intended local wall-clock reading (a `dayjs.utc(getLocalDateString(date, tz)).startOf('day')` trick) — for those, `dayjs.utc()` is exactly the right way to read them back, and this is proven, recently-shipped, tested behavior I do not want to disturb. A live/rolling range (or anything else not produced by that trick) is a *genuine* real instant instead, and needs a real timezone conversion, not a literal digit read. There is no way to tell the two apart by inspecting the `Date` value itself — except that every existing trick-encoded value is, by construction, exactly at a clean unit boundary (`00:00:00.000`/`23:59:59.999`), while a genuine live instant essentially never lands on one by coincidence.

## Fix

**`src/components/charts/charts.fillGaps.hooks.ts` only** — no changes to PR #271's preset code.

1. Import the `timezone` dayjs plugin (already used elsewhere in the codebase: `dates.utils.ts`, `timeRange.utils.ts`).
2. Where `from`/`to` are resolved (current lines 69-78), for any bound that isn't going through the already-trusted `relativeTimeString` → `getRange(timezone)` path (lines 52-56, unchanged): check whether it already sits on a clean unit boundary read as UTC (`.millisecond() === 0 && .second() === 0` for the "start" side / the `.endOf`-shaped equivalent for the "end" side). If it does, treat it as already in the safe local-digit convention (unchanged behavior — this covers the custom absolute-range picker and any other day-precision-anchored value). If it doesn't, treat it as a genuine real instant and convert it via `dayjs.utc(dayjs(date).tz(timezone).format('YYYY-MM-DDTHH:mm:ss.SSS'))` before using it — this is the missing step that realigns a live/rolling bound with how the record data is actually keyed.
3. Leave record-value parsing (`dayjs.utc(value)`, building `recordsByDate`) untouched — already correct.
4. Leave bucket-walking (`current.startOf(granularity)` / `.add(1, granularity)`) as plain UTC duration arithmetic — once `from`/`to` are correctly normalized, this is timezone-invariant for `hour`/`minute`/`second`, and for `day`/`week`/`month`/`quarter`/`year` it's stepping across whole units of an already-local-anchored value, so no `.tz()` is needed in the walk itself.
5. Add a short code comment stating the underlying contract explicitly (a Date reaching this hook is either "local-digits-as-UTC" from a known preset/picker, or a genuine real instant — and the boundary-alignment check is how we tell them apart), since this is genuinely non-obvious and the whole fix hinges on it.

Out of scope / explicitly not doing: touching `dates.utils.ts`, `timeRange.utils.ts`, or `defaults.DateRanges.constants.ts` — their output is already correct for their current consumer (day-precision display alignment) and I don't want to risk PR #271's tested behavior for a fix that's fully achievable inside `useFillGaps` alone. Also out of scope: the separate, pre-existing performance concern that a date-bound chart's query fetches the *entire* unfiltered dataset regardless of `dateBounds` (confirmed via `sample_results.json`, 731 rows fetched to display 3 days) — real, but unrelated to this bug and not something `useFillGaps` controls.

Also out of scope, per earlier discussion: full gap-filling support for Embeddable's extended/custom granularities (e.g. `quarter_hour`, defined server-side per https://docs.embeddable.com/component-libraries/build-components/extending-types) — `useFillGaps` still bails out for any granularity outside the built-in 8 today. I'm not including a heuristic step-inference fallback for these in this fix, since it's an independent, separate gap (no interval metadata is exposed to the frontend for these at all) and mixing it into this timezone fix would blur two unrelated changes. Flagging it as a follow-up if wanted.

## Verification

- Add `src/components/charts/charts.fillGaps.hooks.test.ts` (none exists today despite ~15 chart components depending on this hook). Cover, using the real sample data pattern as a model:
  - Baseline UTC timezone — no regression.
  - Non-UTC timezone, day-precision preset-resolved `dateBounds` (the PR #271 case) — confirms unchanged behavior.
  - Non-UTC timezone, custom absolute-range `dateBounds` from the calendar picker (day-precision, `relativeTimeString: undefined`) — confirms the boundary-detection heuristic correctly treats it as already-safe and doesn't double-convert.
  - Non-UTC timezone, hour granularity, live/rolling `dateBounds` (reproducing this exact bug with the real `sample_results_2.json` shape) — confirms the fix.
- Run the existing suite for consuming components (`BarChartDefaultPro`, `AreaChartPro`, `LineChartComparison*`, `HeatMapPro`, `PivotTablePro`, etc.) to check for regressions.
- `npm run types:check` and `npm run eslint:check`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved chart gap filling for date ranges in different time zones.
  - Corrected handling of manually selected and rolling time ranges.
  - Ensured hour-based charts consistently fill all expected data points without empty values.

- **Tests**
  - Added coverage for UTC, non-UTC, manually selected, and rolling date-range scenarios.

### DIFF
--- a/src/components/charts/charts.fillGaps.hooks.test.ts
+++ b/src/components/charts/charts.fillGaps.hooks.test.ts
@@ -1,0 +1,157 @@
+import { renderHook } from '@testing-library/react';
+import { describe, it, expect, vi } from 'vitest';
+import dayjs from 'dayjs';
+import utc from 'dayjs/plugin/utc.js';
+import type { Dimension, DataResponse } from '@embeddable.com/core';
+import { useFillGaps } from './charts.fillGaps.hooks';
+import { defaultDateRangeOptions } from '../../theme/defaults/defaults.DateRanges.constants';
+import { getTimeRangeFromDateRange } from '../editors/dates/dates.utils';
+
+dayjs.extend(utc);
+
+const isoKey = (date: Date): string => dayjs.utc(date).toISOString().split('Z')[0]!;
+
+const mockUseTheme = vi.fn();
+
+vi.mock('@embeddable.com/react', () => ({
+  useTheme: () => mockUseTheme(),
+}));
+
+const makeTheme = (timezone?: string) => ({
+  clientContext: { timezone },
+  defaults: { dateRangesOptions: defaultDateRangeOptions },
+});
+
+const makeDimension = (overrides: Record<string, unknown> = {}): Dimension =>
+  ({
+    name: 'daily_listens.listened_date',
+    title: 'Listened Date',
+    nativeType: 'time',
+    __type__: 'dimension',
+    inputs: {
+      granularity: 'hour',
+      ...overrides,
+    },
+  }) as unknown as Dimension;
+
+const makeResults = (data: Record<string, unknown>[]): DataResponse => ({
+  isLoading: false,
+  data,
+});
+
+describe('useFillGaps', () => {
+  it('fills gaps correctly in UTC with no dateBounds (baseline, no regression)', () => {
+    mockUseTheme.mockReturnValue(makeTheme('UTC'));
+
+    const dimension = makeDimension();
+    const results = makeResults([
+      { 'daily_listens.listened_date': '2026-01-01T00:00:00.000', value: 1 },
+      { 'daily_listens.listened_date': '2026-01-01T02:00:00.000', value: 2 },
+    ]);
+
+    const { result } = renderHook(() => useFillGaps({ results, dimension }));
+
+    const keys = result.current.data?.map((r) => r['daily_listens.listened_date']);
+    expect(keys).toEqual([
+      '2026-01-01T00:00:00.000',
+      '2026-01-01T01:00:00.000',
+      '2026-01-01T02:00:00.000',
+    ]);
+  });
+
+  it('respects a day-precision preset dateBounds in a non-UTC timezone (PR #271 case, unchanged)', () => {
+    mockUseTheme.mockReturnValue(makeTheme('America/Los_Angeles'));
+
+    const todayOption = defaultDateRangeOptions.find((opt) => opt.value === 'Today')!;
+    const { from, to } = todayOption.getRange('America/Los_Angeles')!;
+
+    const dimension = makeDimension({
+      granularity: 'hour',
+      dateBounds: { from, to, relativeTimeString: 'Today' },
+    });
+
+    const results = makeResults([
+      {
+        'daily_listens.listened_date': isoKey(
+          dayjs
+            .utc(from as Date)
+            .add(3, 'hour')
+            .toDate(),
+        ),
+      },
+    ]);
+
+    const { result } = renderHook(() => useFillGaps({ results, dimension }));
+
+    const keys = result.current.data?.map((r) => r['daily_listens.listened_date']) ?? [];
+    expect(keys.length).toBe(24);
+    expect(keys[0]).toBe(isoKey(from as Date));
+    expect(keys[23]).toBe(
+      isoKey(
+        dayjs
+          .utc(from as Date)
+          .add(23, 'hour')
+          .toDate(),
+      ),
+    );
+  });
+
+  it('treats a manually-picked absolute range from the custom picker as already-safe (no double-conversion)', () => {
+    mockUseTheme.mockReturnValue(makeTheme('America/Los_Angeles'));
+
+    const pickedRange = getTimeRangeFromDateRange(
+      { from: new Date('2026-06-15T00:00:00.000Z'), to: new Date('2026-06-15T23:59:59.999Z') },
+      'America/Los_Angeles',
+    )!;
+
+    const dimension = makeDimension({
+      granularity: 'hour',
+      dateBounds: pickedRange,
+    });
+
+    const results = makeResults([
+      { 'daily_listens.listened_date': isoKey(pickedRange.from as Date), value: 1 },
+    ]);
+
+    const { result } = renderHook(() => useFillGaps({ results, dimension }));
+
+    const keys = result.current.data?.map((r) => r['daily_listens.listened_date']) ?? [];
+    // 24 hourly buckets spanning the picked local day, anchored exactly at pickedRange.from
+    expect(keys.length).toBe(24);
+    expect(keys[0]).toBe(isoKey(pickedRange.from as Date));
+  });
+
+  it('correctly fills an hour-granularity chart bound to a live/rolling range (regression for the reported bug)', () => {
+    mockUseTheme.mockReturnValue(makeTheme('America/Los_Angeles'));
+
+    // Real shape reproduced from the reported bug: 25 continuous naive-local hourly
+    // records (Cube already converted them to America/Los_Angeles wall-clock time),
+    // bound by a genuine real "last 24 hours" instant pair (PDT is UTC-7 in September,
+    // so the local 05:51:14 boundary is 12:51:14 in real UTC).
+    const data = Array.from({ length: 25 }, (_, i) => {
+      const stamp = dayjs.utc('2026-09-02T05:00:00.000').add(i, 'hour');
+      return { 'speed.timestamp': stamp.toISOString().split('Z')[0], value: i };
+    });
+
+    const dimension = makeDimension({
+      granularity: 'hour',
+      dateBounds: {
+        from: new Date('2026-09-02T12:51:14.000Z'),
+        to: new Date('2026-09-03T12:51:14.999Z'),
+        relativeTimeString: undefined,
+      },
+    });
+
+    const dimensionWithName = { ...dimension, name: 'speed.timestamp' } as unknown as Dimension;
+
+    const { result } = renderHook(() =>
+      useFillGaps({ results: makeResults(data), dimension: dimensionWithName }),
+    );
+
+    const rows = result.current.data ?? [];
+    const nullRows = rows.filter((r) => r.value === undefined);
+
+    expect(rows.length).toBe(25);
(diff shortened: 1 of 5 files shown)


cwbuecheler:
@mad-raccoon - Olivier at Orb has confirmed that this PR works much better for him than the previous attempt - it does what he's expecting. I dug in a lot more with Claude Code and tried to make sure I a) followed what it was doing and b) modified as little code as possible (except adding testing for fillgaps, which we should probably have anyway). Can you take a look when you have a sec?

src/components/charts/charts.fillGaps.hooks.ts
```diff
@@ -11,12 +12,49 @@ import { defaultGranularitySelectFieldOptions } from '../../theme/defaults/defau
...
+    value.minute() === 59 &&
+    value.second() === 59 &&
+    value.millisecond() === 999;
+  return isStartOfDay || isEndOfDay;
+};
+
+const resolveBoundary = (rawValue: unknown, tz?: string): dayjs.Dayjs => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const asUtc = dayjs.utc(rawValue as any);
+  const isGenuineInstant = rawValue instanceof Date && tz && !isCleanBoundary(asUtc);
```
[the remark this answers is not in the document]
cwbuecheler: I asked for more explanation here as to why Claude did what it did. The reply:

---

What it's pointing at: In the fix I added last time, isCleanBoundary decides whether a date bound is "already safe" (built from a preset/picker, needs no conversion) versus "a live instant" (needs a timezone conversion) by checking whether the time is exactly 00:00:00.000 or 23:59:59.999. The reasoning was: presets always land exactly on those boundaries, and a genuinely live "now"-based value essentially never does by coincidence.

Why CodeRabbit is flagging it: "Essentially never" isn't "never." If someone's live/rolling range happens to compute its to at exactly midnight (say, a rolling window whose start time was itself exactly midnight and rolls in day-sized increments), my heuristic would misread that live bound as "already safe" and skip the timezone conversion it actually needs — silently reintroducing the same bug for that one edge case. CodeRabbit's suggestion is: instead of guessing which kind of bound this is by inspecting its shape, the code that resolves dateBounds should explicitly know and say whether a given bound came from a preset/custom-picker (safe) or a live/rolling source (needs conversion) — and act on that explicit flag rather than an inference.

Is it a real gap? Yes, it's the same caveat I flagged myself in the plan ("an edge case where a genuine live bound coincidentally lands exactly on a day boundary would be mis-classified... exceedingly rare and low-stakes"). CodeRabbit is right that it's a real theoretical hole; the question is whether it's worth closing given the cost.

Why it's not a small fix: Making this explicit means tracing every place a dateBounds/externalDateBounds value gets constructed — the built-in presets, the custom range picker, and whatever produces the live/rolling values on the customer's side — and having each one tag its output as origin: 'preset' | 'live' (or similar), then threading that tag through dimension.inputs.dateBounds/externalDateBounds all the way into useFillGaps. That touches the TimeRange type itself and multiple files outside this hook, not just charts.fillGaps.hooks.ts — a meaningfully bigger change than the one I made.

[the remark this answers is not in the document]
cwbuecheler: imo we should leave this and not do the increased code work. My reasoning is that it's likely most users' traffic dips at midnight anyway, so the odds of there being a tick at exactly that millisecond are slim, and even if it does happen, it'd be a single datapoint in an ocean, and probably not statistically significant to most clients.

mad-raccoon: I want to push back on the resolution of the CodeRabbit thread above. The failure mode isn't "a datapoint at midnight" — it's a **bound** at midnight.

A host app that builds its own range as `startOfDay(UTC) → endOfDay(UTC)` will hit `00:00:00.000Z` / `23:59:59.999Z` **every time**. That's not a coincidence, it's systematic. When it happens the heuristic skips conversion, and in `America/Los_Angeles` the whole chart shifts by 7 hours — the exact bug this PR is fixing, just for a different customer.

Two ways forward, cheapest first:

1. **Fix at the source.** Have `getTimeRangeFromDateRange` and the presets emit real instants (they already know `tz`), then convert *every* `Date` here via `.tz(tz)` — no guessing. Yes, it touches #271, but #271's "local-digits-as-UTC" encoding is what created the ambiguity in the first place. Every future consumer of `TimeRange` will otherwise have to re-invent this heuristic.
2. **If that's too big right now:** keep the heuristic but be honest about it. Add a test that pins the midnight-UTC live bound as a `// KNOWN LIMITATION`, open a ticket, and rename the PR so nobody reads "Improve time zone handling" as "done".

Either is fine with me. Silently accepting the gap isn't.

src/components/charts/charts.fillGaps.hooks.ts
```diff
@@ -11,12 +12,49 @@ import { defaultGranularitySelectFieldOptions } from '../../theme/defaults/defau
...
+  const isEndOfDay =
+    value.hour() === 23 &&
+    value.minute() === 59 &&
+    value.second() === 59 &&
+    value.millisecond() === 999;
+  return isStartOfDay || isEndOfDay;
+};
+
+const resolveBoundary = (rawValue: unknown, tz?: string): dayjs.Dayjs => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
mad-raccoon: Nit: `rawValue: unknown` + `as any` + eslint-disable. The only inputs here are `Date | string | undefined` (bounds or a raw record value). Tif we type it that way and the cast goes away

src/components/charts/charts.fillGaps.hooks.test.ts
```diff
@@ -0,0 +1,157 @@
...
+      '2026-01-01T02:00:00.000',
+    ]);
+  });
+
+  it('respects a day-precision preset dateBounds in a non-UTC timezone (PR #271 case, unchanged)', () => {
+    mockUseTheme.mockReturnValue(makeTheme('America/Los_Angeles'));
+
+    const todayOption = defaultDateRangeOptions.find((opt) => opt.value === 'Today')!;
+    const { from, to } = todayOption.getRange('America/Los_Angeles')!;
+
```
mad-raccoon, line 87: This test depends on the real clock (`Today` → `new Date()`). It passes today, but it's non-deterministic by design. Wrap with `vi.useFakeTimers()` / `vi.setSystemTime(...)` so it's pinned.

Also missing: a test for the `externalDateBounds` path. That's what `LineChartComparison*` use, and it's what the `useMemo` deps fix touches — right now that fix has no coverage.

mad-raccoon (approved):
1 small comment about types. but approved

mad-raccoon (approved):

PR #274, opened 2026-09-04
TPS-1459: anchor date ranges in target timezone, not UTC

yashrajk17:

This PR is needed because date filters near a timezone boundary could pull in an extra day of data. A date picked for a dashboard in one timezone was still being searched for in UTC time, so it didn't always match the actual day the user picked.

Example: picking "Sept 4" on a dashboard set to Sydney time.

**Before:** searched Sept 4, 12:00am – 11:59pm UTC, which is actually Sept 4, 10:00am – Sept 5, 9:59am in Sydney time — so Sept 5 data was included too.

**After:** searches Sept 4, 12:00am – 11:59pm Sydney time. Only Sept 4 data is included.

# Main changes

- Date ranges are now worked out using the dashboard's own timezone the whole way through, instead of switching back to UTC partway through
- This covers both custom date picks and the built-in presets (Today, Yesterday, Last 7 days, etc.)

# Test evidence

https://github.com/user-attachments/assets/73a50ecb-a982-403f-a8e6-b23117e3f17d

### DIFF
--- a/.changeset/proud-phones-ring.md
+++ b/.changeset/proud-phones-ring.md
@@ -1,0 +1,5 @@
+---
+'@embeddable.com/remarkable-pro': minor
+---
+
+Fix timezone in date component
--- a/src/components/editors/dates/DateRangePickerPresetsPro/DateRangePickerPresetsPro.utils.test.ts
+++ b/src/components/editors/dates/DateRangePickerPresetsPro/DateRangePickerPresetsPro.utils.test.ts
@@ -39,16 +39,22 @@ describe('getDateRangeSelectFieldProOptions', () => {
     });
   });
 
-  it('calls getTimeRangeLabel with the result of getRange and option.dateFormat', () => {
+  it('calls getTimeRangeLabel with the result of getRange, option.dateFormat, and the timezone', () => {
     const returned = range(new Date('2024-01-01'), new Date('2024-01-31'));
     const opt = makeOption('last_7_days', 'key|Last 7 days');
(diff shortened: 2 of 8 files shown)


mad-raccoon:
@yashrajk17 the evidence is not making sense to me

if i select "today" i am not expecting to see a interval of dates - but 1 date only

image

yashrajk17:
> @yashrajk17 the evidence is not making sense to me
> 
> 
> 
> if i select "today" i am not expecting to see a interval of dates - but 1 date only
> 
> 
> 
> image
> 
> 

Sure let me check

yashrajk17:
@mad-raccoon Its fixed now

mad-raccoon:
@yashrajk17 the code looks good, but not sure if this will cover the scenario where the user is using a date without a editor

i.e. i can create a builder variable 'Today' and use it directly on my component - no timezone will be applied

could you double check please ?

yashrajk17:
> @yashrajk17 the code looks good, but not sure if this will cover the scenario where the user is using a date without a editor
> 
> i.e. i can create a builder variable 'Today' and use it directly on my component - no timezone will be applied
> 
> could you double check please ?

yes its working

https://github.com/user-attachments/assets/2dafd39d-897b-4147-a680-2029d873c538

mad-raccoon (approved):

mad-raccoon:
@yashrajk17 we need to 'npx changeset' to create a new version

PR #277, opened 2026-09-09
Y-Axis Other Bucketing

cwbuecheler:
Currently it's possible to bucket the X-axis into "other" buckets, but in the case where the X axis is a date range (or in other cases) the customer may want to bucket the Y-axis instead. This update adds that functionality to multiple charts.

### What this does

Adds a new "Limit top group by" input that ranks the `groupBy` dimension (the one driving side-by-side bars, stack segments, or separate lines) by its total contribution to the measure, keeps the top N groups, and folds everything else into a single "Other" series — independent of any existing x/y-axis limiting. This is the group-dimension analog of the existing `xAxisMaxItems`/`groupTailAsOther` feature, requested by a client who wanted a full date range on X with only the top N products (plus "Other") on Y.

**Charts affected:** `BarChartGroupedPro`, `BarChartGroupedHorizontalPro`, `BarChartStackedPro`, `BarChartStackedHorizontalPro`, `LineChartGroupedPro`.

**New inputs:** `Sort by group total` (`sortDirectionTopGroupBy`) and `Limit top group by` (`limitTopGroupBy`), added to `component.inputs.constants.ts` and each of the 5 charts' `definition.ts`.

### How it works

- **Ranking query** — a server-side query selects `[groupBy, measure]`, sorted/limited to find the top N−1 groups (reserving one slot for "Other," matching `groupTailAsOther`'s existing semantics).
- **Main query** — extended to filter to just the kept top-N groups once the rank query resolves, via the same cache-key/state pattern already used for x-axis top-N ordering.
- **"Other" query** — a groupBy-agnostic grand total per axis bucket (`select:[axis, measure]`, no groupBy filter at all).
- **Subtraction, not an exclusion filter** — "Other" is computed client-side as `grandTotal − sum(kept groups)`. This was a deliberate choice after two rejected alternatives: `notEquals` + an array value has no precedent anywhere in this codebase (the filter-builder UI uses `notContains` for multi-value exclusion instead), and `notContains` does substring matching, so a kept group like `"Widget"` would wrongly swallow an excluded group like `"Widget Pro"` into nowhere. Subtraction avoids both risks entirely, at the cost of restricting the feature to additive measures (sum/count) — enforced by `isOtherBucketableMeasure`, which explicitly excludes `avg`/`min`/`max`/`count_distinct`/`count_distinct_approx` and any unrecognized future `aggType` (allowlist, not blocklist).
- **Shared infrastructure** — `bars.loadData.utils.ts`/`bars.hooks.ts` were promoted to `charts.loadData.utils.ts`/`charts.hooks.ts` (they were already dimension-agnostic) so `LineChartGroupedPro` could reuse the same query/caching machinery the bar charts already had for x-axis ordering. A new `useGroupOtherResults` hook consolidates the per-chart wiring (cache the ranked group order, then merge main + Other) into one call, replacing ~20 duplicated lines per chart with one hook call — addresses a SonarQube duplication flag from an earlier pass.

### Bugs found and fixed along the way

- **Color-collision race**: `results` (main) and `resultsGroupOther` (Other aggregate) are independent async calls that can resolve in either order. If Other's data arrived before main's, a transient render would compute Other alone at color-index 0; since colors are cached by value and never recomputed, a real group later landing at index 0 would collide with it. Fixed by gating Other's computation on both queries having settled, not just one.
- **Stale-data race**: caught by CodeRabbit — the original fix only checked `mainResults.isLoading`, not `resultsGroupOther.isLoading`, so a cache layer that keeps previous data visible during a re-fetch could feed stale grand totals into the subtraction even after `mainResults` settled. Now both must settle.
- Two other CodeRabbit findings were evaluated and applied: scoping the "Other" query to `axisOrder`/`maxResults` like the main query already does, and confirming `isOtherBucketableMeasure` needed to be an allowlist rather than a blocklist.
- Separately, found and fixed a stale `node_modules` issue (`@embeddable.com/remarkable-ui` was at `3.1.1` locally instead of the lockfile's `3.2.0`, missing `chartjs-chart-funnel` entirely) — unrelated to this feature, surfaced as pre-existing `tsc` errors on this branch; not a bad merge, just a local `npm install` that hadn't run since `CS-282: Create FunnelChart Component` landed.

### Testing

Extensive new unit test coverage across `charts.loadData.utils.test.ts`, `charts.utils.test.ts`, and `charts.hooks.test.ts` for the query builders, the subtraction math (including a regression test proving the exact bug `notContains` would have caused, and the stale-data race), plus updated component tests for all 5 charts. Full suite: `tsc --noEmit`, eslint, and prettier all clean; 1442 tests passing.

### Versioning

Changeset added (`add-groupby-other-bucketing.md`, minor bump) describing the feature for the next release.

image

## Summary by CodeRabbit

* **New Features**
  * Added “Limit top group by” controls to grouped bar, stacked bar, and grouped line charts, including horizontal variants.
  * Groups can be sorted by total measure contribution, with remaining values combined into an “Other” series.
  * Supports sum and count measures independently of axis limits.

* **Bug Fixes**
  * Improved “Other” value calculations across axis buckets.
  * Improved grouped chart loading, error handling, and cached group ordering.


cwbuecheler:
@coderabbit review

cwbuecheler:
@mad-raccoon - This is ready for human eyes :) It's a fairly big one. Adds y-axis bucketing to multiple charts (see description showing an example). It touches a lot of files but most of them are just a few lines (tests make up the bulk of the diff).

One thing to note: the Sonar "duplicate code" warning is happening for a specific reason. Here's Claude's explanation:

> Consolidation applied to all 5 charts (4 bar + LineChartGroupedPro), verified: tsc --noEmit, eslint, prettier, and all 1437 tests pass.
>
> What changed: each chart's ~21-line block (call useUpdateGroupOrderAndCacheKey, then separately call mergeGroupOtherResults with 5 positional args) is now a single useGroupOtherResults({...}) call — about 10 lines, passing named options. The actual composed logic (cache the ranked group order, then merge main + Other) now lives once in charts.hooks.ts, tested directly there (3 new tests) rather than only indirectly through each component.
>
> One thing worth knowing: the remaining ~10-line call site is still nearly identical across the 4 bar charts — it differs only in axis: xAxis vs axis: yAxis. That's irreducible without collapsing these into one generic component (which would trade away the real per-chart differences — horizontal vs vertical, grouped vs stacked layout — for questionable benefit), so Sonar may still flag some residual duplication there. But the actual logic duplication is gone; what's left is just each component wiring its own axis into a shared hook, the same shape as the pre-existing useUpdateAxisOrderAndCacheKey call that was never flagged. Worth re-running the Sonar scan to see where the number lands now.

src/components/charts/bars/bars.types.ts
```diff
@@ -19,6 +19,11 @@ export type BarChartStackedBaseProps = {
   axisOrder?: string[];
   axisOrderCacheKey?: string;
   setAxisOrderAndCacheKey?: (values: string[], cacheKey: string) => void;
+  resultsGroupOrder?: DataResponse;
+  resultsGroupOther?: DataResponse;
+  groupOrder?: string[];
```
mad-raccoon: `groupOrder` looks dead. All 5 definitions pass `groupOrder: cachedGroupOrder`, but no `index.tsx` reads it. It's declared here but not in `LineChartGroupedProProp`.

Either use it (order the datasets by rank, like `axisOrder` does in `getBarStackedChartProData`) or drop it from all 5 definitions and this type.

src/components/charts/bars/BarChartGroupedPro/definition.ts
```diff
@@ -135,15 +160,35 @@ const props = (
...
+      dataset: inputs.dataset,
+      limitTopGroupBy: inputs.limitTopGroupBy,
+      groupBy: inputs.groupBy,
+      measure: inputs.measure,
+      sortDirection: groupSortDirection,
+      timezone,
+    }),
+    resultsGroupOther: loadDataResultsGroupOther({
+      dataset: inputs.dataset,
+      axis: inputs.xAxis,
```
mad-raccoon: Nit: axis is passed two ways. Here it's `axis: inputs.xAxis` + `granularity`; in `results` below it's `axis: xAxisWithGranularity`. Same result, but picking one shape would make the 5 definitions easier to scan. (Same in the other 4 charts.)

src/components/charts/charts.loadData.utils.ts
```diff
@@ -0,0 +1,394 @@
...
+  dataset,
+  axis,
+  measure,
+  granularity,
+  groupOrder,
+  axisOrder,
+  maxResults,
+  timezone,
+}: LoadDataResultsGroupOther): DataResponse | undefined => {
+  if (groupOrder == null) return undefined;
```
mad-raccoon, line 298: Extra query when both axis-limit and group-limit are on: `loadDataResults` waits for `axisOrder` before firing, but this one doesn't. So it fires once unfiltered (all axis buckets) as soon as `groupOrder` lands, then again filtered once `axisOrder` arrives.

Not a correctness issue (the both-settled gate in `mergeGroupOtherResults` covers it), just a wasted round-trip. Could gate on `axisOrder` the same way `loadDataResults` does when `shouldGetTopItems` is true.

src/components/charts/charts.loadData.utils.ts
```diff
@@ -0,0 +1,394 @@
...
+// the whole axis range, independent of any axis-order limiting above. Only
+// applies to sum/count measures — avg/min/max can't be correctly represented
+// by a NOT-IN "everything else" aggregate (see isOtherBucketableMeasure).
+
+export const getGroupOrderLimit = (limitTopGroupBy?: number): number | undefined => {
+  const limit = getLimit(limitTopGroupBy);
+  if (limit == null) return undefined;
+  // Reserve one slot for the synthetic "Other" row, matching groupTailAsOther's
+  // maxItems semantics (maxItems total = maxItems - 1 real + 1 Other).
+  return Math.max(limit - 1, 0);
```
mad-raccoon, line 124: `limitTopGroupBy = 1` silently does nothing: this returns 0, so `shouldGetTopGroupItems` is false and the user sees all groups with no "Other". Matches `groupTailAsOther` semantics, so probably intended — but worth saying in the input description (e.g. "minimum 2").

Also: the rank query ranks over the whole dataset, not just the top-N axis buckets. Reasonable, but with axis-limit on, a "top" group can be near-zero in the shown buckets. Maybe a one-line comment here so it's clearly a design choice.

src/components/charts/charts.loadData.utils.ts
```diff
@@ -0,0 +1,394 @@
...
+};
+
+export const getGroupOrderCacheKey = ({
+  dataset,
+  groupBy,
+  measure,
+  sortDirection,
+  limit,
+  timezone,
+}: LoadDataResultsGroupOrderArgs & { limit?: number }): string | undefined => {
```
mad-raccoon: Nit: `LoadDataResultsGroupOrderArgs & { limit?: number }` — `limit` is already in `LoadDataResultsGroupOrderArgs`. The intersection is redundant.

src/components/charts/lines/LineChartGroupedPro/definition.ts
```diff
@@ -156,8 +186,4 @@ export const lineChartGroupedPro = {
     props,
     events,
   },
-  results: {
-    loadDataArgs: loadDataResultsArgs,
-    loadData: loadDataResults,
-  },
 } as const;
```
mad-raccoon, line 187: This drops the `results: { loadDataArgs, loadData }` export. `ARCHITECTURE.md` lists it as part of every definition. The 4 grouped bar charts already lack it, so the 5 are now consistent with each other — but all 5 now deviate from the doc. Either keep a thin `results` export here or update `ARCHITECTURE.md` to cover the shared `charts.loadData.utils.ts` pattern.

mad-raccoon (commented):
Reviewed the shared infra (`charts.loadData.utils.ts`, `charts.utils.ts`, `charts.hooks.ts`) and all 5 chart `definition.ts` / `index.tsx` / tests.

**Consistency:** the 4 bar charts are identical apart from `xAxis`/`yAxis`. `LineChartGroupedPro` matches, minus the axis-order bits it never had. Good.

**Approach:** subtraction for "Other" + allowlist on additive measures is the right call. The both-settled gate in `mergeGroupOtherResults` is correct.

Inline comments:
- dead `groupOrder` prop across all 5 charts (the one worth fixing)
- extra unfiltered `resultsGroupOther` query on first load when axis-limit + group-limit are both on
- `limitTopGroupBy = 1` silently no-ops; rank query ignores axis filter — both fine as design, just make them visible
- nits: axis passed two ways in definitions, redundant type intersection, dropped `results` export vs `ARCHITECTURE.md`

`errorMessage` not including `resultsGroupOther?.error` is fine since the merge folds it into `results.error`.

Happy to approve once `groupOrder` is either used or removed.

cwbuecheler:
@mad-raccoon - All comments resolved. Sonarqube doesn't like the code duplication but as mentioned, Claude has an explanation for that above that I think makes sense.

mad-raccoon (approved):

PR #280, opened 2026-09-14
CS-294: Improvements for Funnel Chart component

miguelopezv:

Solves various bugs/issues with the Funnel Chart component.

# Main changes

- Optimize wording on the tooltip regarding theme variables to avoid confusion.
- Ability to disable the legend but display the category names in/with/next to the slices themselves.
- Legend labels aren't respecting the theme.
- The tooltips only include a number, but not the measure name, nor the Display Name/Prefix/Suffix, etc.
- Clicking the labels inside of the legend pops up an error in the builder and logs a bunch of failures to the console.

## Summary by CodeRabbit

* **New Features**
  * Funnel charts can now display stage labels, value labels, and percentages according to configuration.
  * Legend items support toggling the visibility of individual funnel stages.
  * Tooltips provide formatted count values for clearer chart details.

* **Documentation**
  * Updated color input descriptions to clarify automatic color generation and theme gradient behavior.


src/components/charts/funnel/FunnelChartPro/index.tsx
```diff
@@ -72,6 +80,7 @@ const FunnelChartPro = (props: FunnelChartProProps) => {
         options={options}
         showLegend={showLegend}
         showTooltips={showTooltips}
+        showValueLabels={showValueLabels}
```
mad-raccoon, line 78: this is already on the options right ?

src/components/charts/funnel/funnel.utils.ts
```diff
@@ -75,15 +76,75 @@ const aggregateFunnelStages = (
...
+};
+
 export const getFunnelChartProOptions = (
   theme: Theme = remarkableTheme,
-): Partial<ChartOptions<'funnel'>> => ({
-  plugins: {
-    legend: {
-      position: theme.charts.legendPosition ?? 'bottom',
+  countMeasure: Measure,
+  config: FunnelChartProOptionsConfig = {},
```
mad-raccoon: you dont need to re-type , you already have a type for this - is better to reuse

src/components/charts/funnel/funnel.utils.ts
```diff
@@ -75,15 +76,75 @@ const aggregateFunnelStages = (
...
+          legend.chart.update();
+        },
+        labels: {
+          generateLabels: (chart) => {
+            const colors = chart.data.datasets[0]?.backgroundColor ?? [];
+            const labelColor = chart.options.plugins?.legend?.labels?.color as string | undefined;
+            return (chart.data.labels ?? []).map((label, index) => ({
+              text: String(label ?? ''),
+              fillStyle: (colors as string[])[index],
+              strokeStyle: (colors as string[])[index],
```
mad-raccoon: we can simplify

just create a new const:

```
const color = (colors as string[])[index];

fillStyle: color,
strokeStyle: color
```

src/components/component.inputs.constants.ts
```diff
@@ -312,6 +312,14 @@ const showTotalLabels = {
...
 } as const;
 
+const showStageLabels = {
+  name: 'showStageLabels',
+  type: 'boolean',
+  label: 'Show stage names on slices',
+  defaultValue: false,
+  category: 'Component Settings',
+} as const;
+
```
mad-raccoon: this is specific to a single component, and because of that, should not be reusable - lets define it on the component that uses it - there are multiple examples like this

mad-raccoon: this file and props are exclusively for reusable inputs across components

src/components/charts/funnel/funnel.utils.ts
```diff
@@ -75,15 +76,75 @@ const aggregateFunnelStages = (
...
+
+  if (!config.showStageLabels) return base;
+
+  return {
+    ...base,
+    plugins: {
+      ...base.plugins,
+      datalabels: {
+        display: 'auto',
+        formatter: (value: number, context: Context) => {
```
mad-raccoon: i recommend to create a pure function for this, so we can test separately

```
formatter: getFunnelOptionsDatalabelsFormatter(...)
```

mad-raccoon (changes requested):

src/components/charts/funnel/FunnelChartPro/index.tsx
miguelopezv: Not fully — it's the fallback when showStageLabels is off: FunnelChart builds its own datalabels config then and reads this prop directly (defaulting to true), so removing it would make value labels always show in that mode.

src/components/charts/funnel/funnel.utils.ts
mad-raccoon: export type FunnelChartProProps = {
  stageDimension: Dimension;
  countMeasure: Measure;
  orderDimension?: Dimension;
  startColor?: string;
  endColor?: string;
  results: DataResponse;
  showLegend?: boolean;
  showTooltips?: boolean;
  showValueLabels?: boolean;
  showStageLabels?: boolean;
  displayPercentages?: boolean;
} & ChartCardHeaderProps;


here, this type - we can use it instead of creating another abstraction - single source of truth - if change in the feature, all gets updated - if we have multiple type, there will be breaking change hard to detect

mad-raccoon (changes requested):

src/components/charts/funnel/FunnelChartPro/index.tsx
```diff
@@ -52,7 +55,7 @@ const FunnelChartPro = (props: FunnelChartProProps) => {
   );
 
   const options = mergician(
-    getFunnelChartProOptions(theme),
+    getFunnelChartProOptions(theme, countMeasure, props),
```
mad-raccoon: lets follow the approach we already have in place, the them as the last arg, and the 1st arg as the props needed:

e.g.

```
  const baseOptions = getBarChartProOptions(
    { measures: barMeasures, horizontal: false, data, dimension },
    theme,
  );

```

mad-raccoon (changes requested):
only 1 small improvement.

there are worth sonar comments we should address

mad-raccoon (approved):

